PR #56, opened 2016-02-05
Fix gettext translations

horosgrisa:


bil-elmoussaoui:
Can you only modify the FeedReader.pot? the translations are updated automatically in Transifex

horosgrisa:
@bil-elmoussaoui Check now

bil-elmoussaoui:
Good just need to remove the ru.pot from the commit (and update the translations on Transifex)

jangernert:
oh, already build a new pot file and pushed it to master
so no need to merge this anymore
now you can submit the translations through transifex ;)

PR #57, opened 2016-02-10
Increase max number of articles

horosgrisa:
I use tt-rss with ~3000 articles, that do not appear properly in FeedReader with limit 1000 articles

### DIFF
--- a/src/Widgets/SettingsDialog.vala
+++ b/src/Widgets/SettingsDialog.vala
@@ -117,7 +117,7 @@ public class FeedReader.SettingsDialog : Gtk.Dialog {
     {
 		var sync_settings = headline(_("Sync:"));
 
-		var sync_count = new SettingSpin(_("Number of articles"), settings_general, "max-articles", 10, 1000, 10);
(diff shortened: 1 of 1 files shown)


jangernert:
What exactly do you mean by "3000" articles? You have 3000 articles in the your ttrss database? Or every time you sync, ttrss has collected 3000 new articles that now need to be synced to FeedReader?

horosgrisa:
3000 articles in tt-rss database. But, in FeedReader I see only 1000 older articles.

jangernert:
Then increasing the article count on EVERY snyc to 3000 is a bit over the top. The initial sync after you log in is supposed to download all starred, tagged and unread articles that are currently in the database. Unmodified already read articles are not downloaded as many people made a strong argument to have a faster init sync (can take extremely long when ALL articles are downloaded with the content-grabber) because unstarred, untagged, read articles are not important to have in the local database.
Can't make everyone happy at the same time =/

horosgrisa:
I understand, but by default you have 400 articles. I would like to increase maximal posible number of articles. But rest of users can use default value.

Issue #65, opened 2016-03-09
ArticleView border appears transparent with some gtk themes


abhinandh:
Hi, I tried FeedReader on XFCE with greybird gtk theme. The background for the webkit window appears transparent and it make the content unreadable. Attached is a screenshot to show this in action.
![screenshot_2016-03-09_00-53-16](https://cloud.githubusercontent.com/assets/127667/13630433/3220204a-e594-11e5-82ff-93ed3827f0c2.png)

PR #66, opened 2016-03-09
Fix transparent article view background on some gtk themes.

abhinandh:
#65 Don't set webkit background if background alpha is transparent.


jangernert:
thx

jangernert:
btw: I set the background color primarily for dark themes where the white default background would be way to bright. So what happens now when there is a dark theme that also sets this specific color I grab from the theme to transparent? Is there another widget you can grab the bg-color from with themes like greybird? Would be the cleaner solution ;)
but thx anyway for your commit =)

PR #126, opened 2016-05-25
Fix typo in README

supervacuo:
Setup instructions use `&` which causes `cd ./build` to be run before `mkdir build` has completed, leading to an error. 

`&&` gives the expected behaviour.


jangernert:
Thx for noticing ;)

PR #129, opened 2016-05-28
FIX Close button on 'About' dialog was not working

lcapalbo:
About window remained open after clicking Close button.


jangernert:
Thank you very much =)

PR #175, opened 2016-08-05
Fix up appdata file id to match with the desktop file

petefwalter:
The two need to match for this to work.


jangernert:
Thx =)
I just copied some example file when I originally created the appdata file :D

PR #206, opened 2016-09-15
Theoldreader API

kanishkablack:
TheOldReader API Integeration

Needs testing on few functions

----------
List of Bugs
----------
- [x] marking all articles of a feed as read does not work ( time stamp in nano seconds )
- [x] same goes for categories ( time stamp in nano seconds )
- [x] lots of unneeded error messages from send_request()
- [x] got a crash after initializing m_utils correctly because the setting "user-email" can't be found (I think you confused "username" with it)
- [x] m_utils of theoldreaderAPI never gets initialized

-------
Edit List of bug reported


jangernert:
Can you move the icons(feed-service-theoldreader) directly to the plugin-folder like it is done for the rest of the plugins?
That way all the data for the back-end is located in one folder.

kanishkablack:
Icons are not yet done, i am not a graphic designer

jangernert:
I did already tell you in your last pull-request that I added [oldreader-icons](https://github.com/jangernert/FeedReader/tree/master/data/icons/hicolor/64x64/places) !?!

> I think I fixed the crash. Can you confirm on your PC?
I also drew some old-reader icons in Inkscape and added them to the repository.
Not sure if feedhq even has an official logo.

--------------

> - crash is fixed.
- Icons are updated
- FeedHQ doesn't have any logo yet.

jangernert:
You also need to adjust this line with the correct icon-name:
https://github.com/jangernert/FeedReader/pull/206/files#diff-64a8f7cd60551604f35dddf7d5dd18eeR44

kanishkablack:
@jangernert crash is not fixed

jangernert:
"crash" ... so if I would know what you are talking about I could do something about it...
...
...

kanishkablack:
Some local cache creating the error.
I guess this plugin is up for testing

jangernert:
Local cache? What?

kanishkablack:
i tried to run installed version rather than a dev build, that cased error, not an issue in the code base.

jangernert:
Just tested it. Haven't tried everything yet, but here is what I've noticed:
* no feed-icons
* marking all articles of a feed as read does not work
* same goes for categories
* lots of unneeded error messages from send_request()
* m_utils of theoldreaderAPI never gets initialized
* do we really need "the" in front of everything? makes the names pretty long

Have you tested all the functionality yourself?

jangernert:
Regarding m_utils, you are missing this line in the constructor:
```
m_utils = new TheOldReaderUtils();
```

In the utils the functions getEmail() and setEmail() are not needed. Instead you should use setUser() [here](https://github.com/jangernert/FeedReader/pull/206/files#diff-bb48584d88c5f2281c79c5468df69dc1R76). This will prevent the crash, since there is no "user-email" setting in your settings-schema.

Also I changed the function getFeedsAndCats() of the interface to return a boolean. Return true if everything went fine and false if some sort of error occurred ;)

kanishkablack:
- marking all articles of a feed as read does not work ( fixed )
- same goes for categories (fixed)
- lots of unneeded error messages from send_request() (fixed)

- do we really need "the" in front of everything? makes the names pretty long

I am not much of nomenclature guy, i just kept it.

- got a crash after initializing m_utils correctly because the setting "user-email" can't be found (I think you confused "username" with it)

Regarding that few services treats username as email( like feedly ) other don't ( like inoreader and bazqux)

jangernert:
> Regarding that few services treats username as email( like feedly ) other don't ( like inoreader and bazqux)

We just need something that represents the account of the user. Some services provide the real name, others a username and other services just have an email-address. So just take the property that makes the most sense and is available.

kanishkablack:
bool getFeedsAndCats migrated

jangernert:
Okay, so what is it that you return in accountName() ? The email or the username?
If you return the username then I don't think the email is needed.

kanishkablack:
Username will be returned,
Yep email is not needed

jangernert:
Well, okay I guess it doesn't hurt when we get the email of the user even though it's not needed.
Have you tested adding, removing and renaming feeds and categories and creating new tags and tagging articles?
If so, and everything works, the last thing you have to do is add your name to the end of AboutInfo.authors in Constants.vala.in

kanishkablack:
tested adding, removing and renaming feeds and categories
Tags are not supported ( no testing)

PR #211, opened 2016-09-25
Update recommendedFeeds.json

jscurtu:
Added Lifehacker, a generic RSS feed that I think may be intressted: Lifehacker is a weblog about life hacks and software.


jangernert:
Thx :)
I was hoping you would also add a little description. I was just filling in "bla bla bla" as a placeholder while developing the actual code^^

jscurtu:
Yea I though I can be lazy there too :-) 
I updated it though now.

PR #215, opened 2016-09-26
Solus is just called Solus, not Solus OS

quassy:
Nit-pick: It's simply Solus. (I believe the history is Solus OS -> Evolve OS -> Solus)


jangernert:
The more you know ...

PR #255, opened 2016-11-02
fix build instructions for Pocket plugin

dsboger:


jangernert:
Looks good. If the definition of GOA_API_IS_SUBJECT_TO_CHANGE is going into the cmake-file of every plugin that uses it, I guess it's no longer needed in the root-file. But I will change that later.
```
add_definitions(${POCKET_DEPS_CFLAGS} -DGOA_API_IS_SUBJECT_TO_CHANGE)
```

Thx for your efforts =)

PR #281, opened 2016-12-13
Call m_window.present() in FeedReader.activate(), fix #280

### LINKED ISSUE #280: Missing call to m_window.present() on FeedReader.activate()
Method activate() of FeedReader calls `m_window.show_all()`, but not `m_window.present()`. This causes the issue of not bringing the window upfront when launching the app if it is already running (in GNOME Shell, at least). The situation that led me to investigate was that clicking a notification does not bring the window up if FeedReader is already running. I can confirm that adding the `m_window.present()` fixes this.

dsboger:
Sure!

### DIFF
--- a/src/FeedReader.vala
+++ b/src/FeedReader.vala
@@ -67,6 +67,7 @@ namespace FeedReader {
 			}
 
 			m_window.show_all();
+			m_window.present();
 
 			try
 			{


bil-elmoussaoui:
Thank you!

PR #311, opened 2017-01-05
Create rebuild

fbruetting:
Because doing just `make && make install` without `cmake ..` sometimes causes issues.

### DIFF
--- a/rebuild
+++ b/rebuild
@@ -1,0 +1,2 @@
+#! /usr/bin/sh
+(cd build && rm -rf * && cmake .. && make -j`nproc --all` && sudo make install)


fbruetting:
Still needs `chmod +x`.

fbruetting:
Executing just one command instead of five.

bil-elmoussaoui:
Building from source code is not really recommended until you know what are you doing. As the source code might be unstable (not working as expected). You can add the rebuild command as an alias and just type `rebuild` in your terminal. 
@jangernert Do you think this should be included?

fbruetting:
Ok, an alias is also a good option. :)

bil-elmoussaoui:
The alias should be configured by the user, and it's not up to FeedReader to do so! You can create a small bash script that installs the latest git version for you.

fbruetting:
Yes, I know what an alias is.

jangernert:
I don't think this should be in the repo.
This script just compiles with default settings and paths, no way to pass CMake arguments.
If someone wants to automate the build there are better alternatives.

PR #356, opened 2017-02-08
Flatpak fixes

hadess:
A couple of fixes I used to create a Flatpak for testing your latest fixes.

### DIFF
--- a/flatpak/org.gnome.FeedReader.json
+++ b/flatpak/org.gnome.FeedReader.json
@@ -14,7 +14,9 @@
         /* Access to X11 and XShm (video playback) */
         "--socket=x11",
         "--share=ipc",
+        /* Access to Wayland */
         "--socket=wayland",
+        /* Access to Network */
         "--share=network",
         /* Needed for dconf to work */
         "--filesystem=xdg-run/dconf", "--filesystem=~/.config/dconf:ro",
--- a/flatpak/org.gnome.FeedReader.json
+++ b/flatpak/org.gnome.FeedReader.json
@@ -11,9 +11,9 @@
     "finish-args": [
         /* Play sounds */
         "--socket=pulseaudio",
-        /* Access to feedreader configuration and cache files */
-        "--share=ipc",
+        /* Access to X11 and XShm (video playback) */
         "--socket=x11",
+        "--share=ipc",
         "--socket=wayland",
         "--share=network",
         /* Needed for dconf to work */
--- a/flatpak/org.gnome.FeedReader.json
+++ b/flatpak/org.gnome.FeedReader.json
@@ -16,6 +16,8 @@
         "--share=ipc",
         /* Access to Wayland */
         "--socket=wayland",
+        /* Access to OpenGL */
+        "--device=dri",
         /* Access to Network */
         "--share=network",
         /* Needed for dconf to work */
--- a/README.md
+++ b/README.md
@@ -102,3 +102,4 @@ make
 sudo make install
 ```
 Arch users can build the latest version using `yaourt -S feedreader-git`
+You can also create a Flatpak bundle using `./build_flatpak.sh`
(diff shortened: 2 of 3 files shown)


flatpak/org.gnome.FeedReader.json
```diff
@@ -16,6 +16,8 @@
         "--share=ipc",
         /* Access to Wayland */
         "--socket=wayland",
+        /* Access to OpenGL */
+        "--device=dri",
```
jscurtu: Why?

jscurtu:
Why allow access to DRI, can you explain this?
A script already exist, its the make file ..

hadess:
> Why allow access to DRI, can you explain this?

It's in the commit message.

> A script already exist, its the make file ..

There's no "Makefile" in the repository, there's a CMakeLists.txt file, which needs to be processed, to create the Makefile, to then allow creating the Flatpak.

Would be nice not to close the PR before I have a chance to answer your questions.

jscurtu:
The make file is here:
https://github.com/jangernert/FeedReader/tree/master/flatpak

It does not need to be processed after Cmake.

hadess:
The other changes still apply, and would stop this error:
libEGL warning: wayland-egl: could not open /dev/dri/card0 (No such file or directory)
when running under Wayland, and allow WebKit to use GL for compositing.

hadess:
Actually, it is WebKit creating a Wayland compositor that throws this error:
```
libEGL warning: wayland-egl: could not open /dev/dri/card0 (No such file or directory)
WaylandCompositor requires eglCreateImage and eglDestroyImage.
Nested Wayland compositor could not initialize EGL
```

hadess:
Also, there are bugs or misunderstandings in the Makefile.
```
BRANCH    = master
VERSION   = $(BRANCH)-` date +"%Y%m%d" `

dist:
       flatpak build-bundle repo/ $(BUNDLE).flatpak org.gnome.$(PACKAGE) $(BRANCH)
```

I doubt that's the intended behaviour. The BRANCH above I guess is supposed to be the git branch. But that's not used anywhere, and the json file will always point to the git master branch. The BRANCH below, in the `flatpak build-bundle` command is an ostree branch. It's always master if you don't use branches in ostree.

hadess:
I've updated the branch to take the location of the Makefile into account. Do I need to open a new PR, or could this one be reopened?

jscurtu:
you have to create a new pull request as the branch was repushed or recreated.

jscurtu:
I am not completely satisfied with the Makefile though.. But the base is ok.. 
The way I see it, if is not x86_64 will still create the filename with "x86_64" even if 32bit detected. 
Also, even though there is a $branch you use static master on creating the bundle.

jscurtu:
There are more things that need to be discussed, why was the branch removed from the builder json file? I think it should stay there.. 
If you cant properly sed it and replace it in the Makefile then it needs to be rethink'ed.

Or is there a different reason?

jscurtu:
after looking what other webkit apps like GNOME Web do, access to dri seems legit and accepted.

jscurtu:
Please create a new PR for: 

flatpak: Allow access to OpenGL 
flatpak: Add comments for Wayland and network permissions

and I will accept those...

The others need a little more work, which is why it was so basic in the first place:
- rework the logic of the Makefile 

The following is not very well placed and formatted it also needs to be redone. 
- README: Add instructions on how to build Flatpak bundle

hadess:
> you have to create a new pull request as the branch was repushed or recreated

If this PR was still opened, it would automatically be updated with my updated branch.

hadess:
> The way I see it, if is not x86_64 will still create the filename with "x86_64" even if 32bit detected.

I don't see how it would do that.

Anyway, I'll wait until you reopen this to comment further.

bil-elmoussaoui:
@hadess We canno't reopen the PR, github shows an insensitive button with "the flatpak-fixes-1 branch was forced-pushed or recreated"

PR #359, opened 2017-02-09
Flatpak fixes 1

hadess:


hadess:
@jscurtu shell in a Makefile is awful, but at least it's functional :/

jangernert:
Regarding the line:

> You can also create a Flatpak bundle running `make` in the `flatpak/` sub-directory.

Don't you also have to do a `make dist`. So shouldn't the 'all'-target also execute the 'dist'-target?

hadess:
> Don't you also have to do a make dist. So shouldn't the 'all'-target also execute the 'dist'-target?

No, you're right, fixed in the updated patch set.

PR #375, opened 2017-02-15
Use Gtk.show_uri_on_window() instead of Use Gtk.show_uri()

jscurtu:


jscurtu:
@jangernert please review if you are ok with the changes..

jangernert:
Looks okay. But in the mainwindow-class a simple "this" instead of "this.get_default()" should be enough ;)

jscurtu:
Ok, will change.. thx

PR #388, opened 2017-02-18
Fix some string capitalisation in the UI

pwithnall:
Trivial patches to make some bits of the UI conform to the [GNOME HIG](https://developer.gnome.org/hig/unstable/writing-style.html.en#capitalization).


jangernert:
Thx for the patch

PR #399, opened 2017-02-22
share to Telegram plugin

jscurtu:
- Tested with original Telegram Desktop client


jangernert:
Awesome stuff. That came out of nowhere.
I just looked through the code and saw that there is also a text-property that can be sent to telegram.
Would be cool to be able to write that in FeedReader. Are you familiar with Gtk+? If so you could take a look at the TwitterForm and adapt it to Telegram. If not I will add it after the merge.
I'm a bit busy the next few days. But I try to test the code as soon as possible :)
Or maybe @bil-elmoussaoui can also confirm everything is okay and merge the code.

jscurtu:
On my mobile and mac I send often links to telegram, this was something I was really missing on FeedReader. 
I know there is a text property :-). But, now that I see the TwitterForm, your right that would be a great fit.. I can adapt that to it (tomorrow).. 
I dont think there is no way to detect if Telegram is installed, its in a try catch and will puff to no where... But it does open telegram if it is not running..

jangernert:
That's what I meant by non-functional. There is a telegram-icon, a user clicks on the icon and *poof* nothing happens -> confused user.

jscurtu:
can we add it to the Settings->Share dialog so the user has to explicit add it?

jscurtu:
ok :-)

jangernert:
I think that would need some soft of "single instance" property of the share-accounts, as you wouldn't want to add more than one telegram-buttons. But I can add that easily.

jscurtu:
ok i almost have it .. but i need some sleep first .. :-)
Will continue later... 

bildschirmfoto vom 2017-02-22 05-12-38
bildschirmfoto vom 2017-02-22 05-12-13
bildschirmfoto vom 2017-02-22 05-08-21

jscurtu:
@jangernert @bil-elmoussaoui ready for review.

jscurtu:
also, I rethought the widget for the text field. I didnt like the extra step so I left it out. 
It will open telegram anyways and ask you to choose a reciept. When choosen then you can just type your text if needed before you send.

bil-elmoussaoui:
I will test it right now

bil-elmoussaoui:
All works fine! For me this one is ready to be merged :)
Still not sure about not providing a TextArea for the user to type the message, this is much easier than having to type the text later :p 
@jangernert What do you think about that?

bil-elmoussaoui:
Ping me once it's done so I can give it a try and merge it !

jscurtu:
@bil-elmoussaoui i am done :-)

bil-elmoussaoui:
Great work ! :D

jangernert:
Great work :)
I'll build and test the code later tonight. I noticed a few smaller things I'll clean up and maybe simplify the whole account-ID thing, since there should only be one account.

jscurtu:
Thx. I thought it was more like "escape characters" in string (like C#).. But nice to know, thanks..

PR #414, opened 2017-02-27
implement font settings

bil-elmoussaoui:
screenshot from 2017-02-27 15-57-48

Here you go! 
I've got only a small problem:  It's impossible to disable font size in Gtk FontButton, which means that the output of Gtk.FontButton is always "FontName FontSize" and i don't really know how to handle that except splitting the text by space and removing the latest element from the Array and joining the text again?
screenshot from 2017-02-27 15-59-15

### DIFF
--- a/src/Widgets/AddPopover.vala
+++ b/src/Widgets/AddPopover.vala
@@ -62,13 +62,14 @@ public class FeedReader.AddPopover : Gtk.Popover {
 		m_feedGrid.attach(m_catEntry, 1, 1, 1, 1);
 		m_feedGrid.attach(addButton, 0, 2, 2, 1);
 
-		var opmlLabel = new Gtk.Label(_("OPML-File:"));
+		var opmlLabel = new Gtk.Label(_("OPML File:"));
 		opmlLabel.expand = true;
-		m_chooser = new Gtk.FileChooserButton(_("Select OPML-file"), Gtk.FileChooserAction.OPEN);
+		m_chooser = new Gtk.FileChooserButton(_("Select OPML File"), Gtk.FileChooserAction.OPEN);
 		var filter = new Gtk.FileFilter();
 		filter.add_mime_type("text/x-opml");
 		m_chooser.set_filter(filter);
-		m_chooser.expand = true;
+		m_chooser.expand = false;
+		m_chooser.get_style_context().add_class("flat");
 
 		var importButton = new Gtk.Button.with_label(_("Import"));
 		importButton.get_style_context().add_class("suggested-action");
--- a/src/Widgets/AddPopover.vala
+++ b/src/Widgets/AddPopover.vala
@@ -68,8 +68,7 @@ public class FeedReader.AddPopover : Gtk.Popover {
 		var filter = new Gtk.FileFilter();
 		filter.add_mime_type("text/x-opml");
 		m_chooser.set_filter(filter);
-		m_chooser.expand = false;
-		m_chooser.get_style_context().add_class("flat");
+		m_chooser.expand = true;
 
 		var importButton = new Gtk.Button.with_label(_("Import"));
 		importButton.get_style_context().add_class("suggested-action");
--- a/src/Widgets/SettingsDialog.vala
+++ b/src/Widgets/SettingsDialog.vala
@@ -18,6 +18,7 @@ public class FeedReader.SettingsDialog : Gtk.Dialog {
     private Gtk.ListBox m_serviceList;
     private Gtk.Stack m_stack;
     private InfoBar m_errorBar;
+    private Gtk.HeaderBar headerbar;
     private static SettingsDialog? m_dialog = null;
 
 	public static SettingsDialog get_default()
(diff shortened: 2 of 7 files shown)


src/Widgets/SettingsDialog.vala
```diff
@@ -18,6 +18,7 @@ public class FeedReader.SettingsDialog : Gtk.Dialog {
     private Gtk.ListBox m_serviceList;
     private Gtk.Stack m_stack;
     private InfoBar m_errorBar;
+    private Gtk.HeaderBar headerbar;
```
jangernert: member variable should always start with "m_"

src/Widgets/AddPopover.vala
```diff
@@ -62,13 +62,14 @@ public class FeedReader.AddPopover : Gtk.Popover {
...
+		var opmlLabel = new Gtk.Label(_("OPML File:"));
 		opmlLabel.expand = true;
-		m_chooser = new Gtk.FileChooserButton(_("Select OPML-file"), Gtk.FileChooserAction.OPEN);
+		m_chooser = new Gtk.FileChooserButton(_("Select OPML File"), Gtk.FileChooserAction.OPEN);
 		var filter = new Gtk.FileFilter();
 		filter.add_mime_type("text/x-opml");
 		m_chooser.set_filter(filter);
-		m_chooser.expand = true;
+		m_chooser.expand = false;
+		m_chooser.get_style_context().add_class("flat");
```
jangernert: why exactly was this change made?
The text feels off-positioned now, and I'm not a big fan of making everything flat.

bil-elmoussaoui: It's just the button looks so big in smaller windows (i will revert that)

src/Widgets/SettingsDialog.vala
bil-elmoussaoui: @jangernert I will get that done right now :)

bil-elmoussaoui:
@jangernert Alright, so for you it's better to remove the old font size that we used to have? and just using Gtk.FontButton for everything? 
Seems to be a good idea for me. 

And what do you think about splitting the font size from the font familly ?

jangernert:
Yes, like I said, I think it's better to use the FontButton for both type and size.

I don't really understand what you mean with the splitting. Are you talking about the displayed string on the button itself or the return value from the button after the font has been changed? In that case you can create a FontDescription object with the from_string constructor and query all needed information from that object. Just look at the description over here:
https://valadoc.org/gtk+-3.0/Gtk.FontButton.get_font_name.html

bil-elmoussaoui:
@jangernert My bad didn't saw the description on the Valadocs! i will get that ready in few minutes :)

jangernert:
@bil-elmoussaoui I think I finished the vala-part of the code. There is only a single font gsetting left that saves the whole font-string of the FontButton. The string is parsed inside the buildArticle() function.

bil-elmoussaoui:
@jangernert Please check my comment on your latest commit ;-)

jangernert:
Fixed.

jangernert:
Only thing left is the fontsize.
The size property of the description is a bit weird right now
```
[ DEBUG ] URW Chancery L Medium Italic 12
[ DEBUG ] fontsize: 12288
```

bil-elmoussaoui:
@jangernert Because the font size is not absolute, check the documentation, i'm trying to figure out how to convert it to pixels.

bil-elmoussaoui:
@jangernert + I was thinking about removing the font family defined on the css file so the one chosen by the user can override the rest?

jangernert:
Okay, but we should keep the monospace font for code :)
Regarding the fontsize: I know, it should be in points, right? I checked get_size_is_absolute() and it returned false.

bil-elmoussaoui:
Yeah i kept the monospace for code :)I got the same result

jangernert:
Did a quick workaround.

bil-elmoussaoui:
I got a better solution :) I will push it just right now

jangernert:
Don't you need the "pt" at the end to make it work?

jangernert:
nvm

bil-elmoussaoui:
Can you test it and see if everything works as excepted in your case? The log output shows the right size here :)

jangernert:
Size seems a bit too small over here. I had it at 12pt. Now 12pt is tiny and I have to go up to 16pt to have roughly the same size.

bil-elmoussaoui:
@jangernert Well, we can change the default font size; and it's up to the user to decide which default font settings they want?

jangernert:
That's not the point. Shouldn't be 12pt the same size independently from the application. With the latest commit 12pt in libreoffice is now the same size as 16pt in FeedReader.
Are you sure nothing went wrong when you convert to px? Also why not convert to pt?

jangernert:
> When setting font sizes, device units are always considered to be points (as in "12 point font"), rather than pixels.

jangernert:
Now everything seems to be in order.

bil-elmoussaoui:
Great work on this ;) I will open a new PR for the articles theming tomorrow

jangernert:
Well, most of the work was done by you sir :D

PR #416, opened 2017-02-28
Font settings improvements

bil-elmoussaoui:
- Don't hard code font size 
   I used a large font size and a small font size, the larger size is the selected size + 2pt and the small size is the selected one -2pt, this way the font size won't be too small if you have a  4K screen and you wanted to have larger fonts;

- Follow font style: (not done yet)
For now, if the selected font is bold/italic or whatever kind of font, it's won't work as expected as we don't implement this yet. Should we really implement this or not?

### DIFF
--- a/data/ArticleView/article.html
+++ b/data/ArticleView/article.html
@@ -13,9 +13,9 @@
 <body style="font-size:$FONTSIZEpt;font-family:$FONTFAMILY;">
     <div class="$THEME">
         <header class="post unselectable">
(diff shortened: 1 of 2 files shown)


jangernert:
I thought about the font-style as well. But I don't think it's a good idea. Some articles do have italic or bold text in them. The author intended for this text to look different than the rest of the text, be it a quote or some highlighted paragraph. If the text is already italic, there is no way to make it "more italic" for these paragraphs :D

bil-elmoussaoui:
Perfect, i thought about the same thing

PR #445, opened 2017-04-09
Feedbin improvements

brendanlong:
Some general Feedbin improvements. This gets icons downloading in a reasonable amount of time, but there's still more work that needs to be done (respecting download limits and having the grabber finish in a reasonable amount of time).


brendanlong:
I have a fix for respecting the download limits but I haven't been able to test it yet since I still haven't had a full download complete yet.

jangernert:
Nice work so far =)
Returning the status-code is a neat idea. I should do this for all the other back-ends as well.
The only other back-end using Utils.downloadIcon() is the local one if I remember correctly. So making use of your new method there would eliminate the need to keep the old one for combatibility and you could shorten the name again.

PR #447, opened 2017-04-10
Use Gee.List interface instead of requiring exactly Gee.LinkedList

brendanlong:
This just changes all references to `Gee.LinkedList` in type parameters to `Gee.List`. Everything has the same type as before, but it will be a lot easier to use certain functions (like `List.slice`) if we use the interface instead of a concrete type.

The only change needed was to reverse the list in `FeedServer.writeArticles()` in a way that doesn't require a bidirectional iterator, and hopefully this is easier to read anyway.

### DIFF
--- a/plugins/backend/bazqux/bazquxAPI.vala
+++ b/plugins/backend/bazqux/bazquxAPI.vala
@@ -80,7 +80,7 @@ public class FeedReader.bazquxAPI : GLib.Object {
 		return false;
 	}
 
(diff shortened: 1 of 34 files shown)


jangernert:
Looks solid.
Like I said, maybe it's a better idea not to reverse the list and copying ever element in the process. But to just iterate through the list backwards in the db-write-funtion. But that is something that can be added later on.

PR #448, opened 2017-04-11
Add tooltips for two tool buttons

manuq:
Adding tooltips for the following tool buttons that are missing one:

- Star icon to mark article as starred / unstarred
- Circle icon to mark article as read / unread

This assists users with disabilities as well as giving a translation for
an unrecognized icon, like the circle for Mark read/unread that isn't
obvious.

### DIFF
--- a/src/Widgets/ArticleViewHeader.vala
+++ b/src/Widgets/ArticleViewHeader.vala
@@ -42,11 +42,13 @@ public class FeedReader.ArticleViewHeader : Gtk.HeaderBar {
 
         m_mark_button = new HoverButton(unmarked_icon, marked_icon, false);
 		m_mark_button.sensitive = false;
+		m_mark_button.set_tooltip_text(_("Mark article as starred or unstarred"));
 		m_mark_button.clicked.connect(() => {
 			toggledMarked();
 		});
 		m_read_button = new HoverButton(read_icon, unread_icon, false);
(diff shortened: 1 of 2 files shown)


jangernert:
The buttons had tooltips in earlier versions. But this caused bugs with the enter- and leave-events of the buttons.

brendanlong:
I wonder if we should open a bug against GTK to get that fixed?

manuq:
@jangernert I see the on-enter on-leave events are being catched to change the tooltip icon. Won't you think is better to fix that implementation instead? Would you accept a patch for that?

manuq:
@jangernert FeedReader.HoverButton seems a custom implementation of Gtk.ToggleToolButton.

jangernert:
I don't think you can fix it in the code of FeedReader. There was always an edge-case that caused it to misbehave with tooltips. If you find some magic combination of event-attributes that works for all cases, I'm happy to merge the changes.
I don't know how exactly ToggleToolButton works, but as far as I know there is no gtk-widget that behaves like HoverButton.

manuq:
@jangernert I think FeedReader should do like Corebird, with a GtkToggleButton styled with CSS to change the icon color when the button is toggled:
corebird-example

If you like it I can work on this. Master is not working for me, failing to sync articles, but when it works I can do it.

PR #451, opened 2017-04-16
refactor favicon code a bit

jangernert:
Favicons are now grabbed in following order:
- check html for `<link rel="shortcut icon">`
- check html `<link rel="icon">`
- check `domain/favicon.ico`
- as a last fall-back use `https://f1.allesedv.com`

Also all the back-end specific download functions are now replaced by one single function. This function does a HEAD-request first and compares etags (saved as txt-file for simplicity). After downloading also does a md5-check so it does not unnecessarily writes all the icons to disc every time (thinking about throwing this one out again).

Anything I missed?


src/Utils.vala
```diff
@@ -427,45 +438,189 @@ public class FeedReader.Utils : GLib.Object {
...
+			return true;
+
+
+		// last chance: try allesedv service
+		string allesedv_url = siteURL;
+		if(allesedv_url.has_prefix("http://"))
+			allesedv_url = allesedv_url.replace("http://", "");
+		else if(allesedv_url.has_prefix("https://"))
+			allesedv_url = allesedv_url.replace("https://", "");
+		allesedv_url = "https://f1.allesedv.com/32/%s".printf(allesedv_url);
```
brendanlong: I'm not sure if this provides much value if we're already doing the other steps.

src/Utils.vala
```diff
@@ -427,45 +438,189 @@ public class FeedReader.Utils : GLib.Object {
...
 				}
-				return true;
 			}
-			Logger.error(@"Error downloading icon for feed: $feed_id");
-			return false;
+			else
+			{
+				Logger.warning(@"HEAD-request failed $icon_url");
+				return false;
+			}
```
brendanlong: Rather than doing a HEAD and checking for etag, we should send the etag with the GET request and check for a 304 response.

src/Utils.vala
```diff
@@ -427,45 +438,189 @@ public class FeedReader.Utils : GLib.Object {
...
+			if(FileUtils.test(local_filename, GLib.FileTest.EXISTS))
+			{
+				var remoteMd5 = GLib.Checksum.compute_for_string(GLib.ChecksumType.MD5, (string)message_dlIcon.response_body.flatten().data);
+				var localMd5 = GLib.Checksum.compute_for_string(GLib.ChecksumType.MD5, getFileContent(local_filename));
+				if(remoteMd5 == localMd5)
+				{
+					// file exists and is identical to remote file
+					// we already downloaded it, but there is no need to write to the disc
+					Logger.debug("Utils.downloadIcon: md5 identical -> icon unchanged");
+					return true;
```
brendanlong: Reading the file and running MD5 is probably more expensive than just writing the new file to disk.

plugins/backend/ttrss/ttrssAPI.vala
```diff
@@ -20,11 +20,26 @@ public class FeedReader.ttrssAPI : GLib.Object {
 	private string m_ttrss_sessionid;
 	private uint64 m_ttrss_apilevel;
 	private Json.Parser m_parser;
+	private string? m_iconDir = null;
```
brendanlong: Why does this backend use a different icon directory than the one all of the others use?

src/Utils.vala
jangernert: I was thinking mostly about SSDs. But like I said, I'm also not sure if it's worth it.

src/Utils.vala
jangernert: Will do

src/Utils.vala
jangernert: I'm not sure if allesedv does anything more than this code does. But yeah, it's probably not helping that much.

src/Utils.vala
```diff
@@ -427,45 +438,189 @@ public class FeedReader.Utils : GLib.Object {
...
+			var xpath = grabberUtils.getURL(doc, "//link[@rel='apple-touch-icon']");
 
-		if(!FileUtils.test(local_filename, GLib.FileTest.EXISTS))
+			if(xpath == null)
+			// check for <link rel="shortcut icon">
+				xpath = grabberUtils.getURL(doc, "//link[@rel='shortcut icon']");
+
+			if(xpath == null)
+				// check for <link rel="icon">
+				xpath = grabberUtils.getURL(doc, "//link[@rel='icon']");
```
jangernert: Didn't know that. Will change

src/Utils.vala
```diff
@@ -427,45 +438,189 @@ public class FeedReader.Utils : GLib.Object {
...
+	        html_cntx.use_options(Html.ParserOption.NOERROR + Html.ParserOption.NOWARNING);
+	        Html.Doc* doc = html_cntx.read_doc(html, siteURL, null, Html.ParserOption.NOERROR + Html.ParserOption.NOWARNING);
+	        if(doc == null)
+	        {
+	            Logger.debug("Utils.downloadFavIcon: parsing html failed");
+	    		return false;
+	    	}
+
+			// check for <link rel="apple-touch-icon">
+			var xpath = grabberUtils.getURL(doc, "//link[@rel='apple-touch-icon']");
```
jangernert: Currently the higher resolution is not of any value to FeedReader. The images are loaded as pixbufs at 24x24 while keeping the aspect ratio. At this point this is not taking HIDPI into account. But I hope at some point this code can also be made HIDPI-aware and the higher resolution icons will be of value.

src/Utils.vala
```diff
@@ -427,45 +438,189 @@ public class FeedReader.Utils : GLib.Object {
...
+		int first = hostname.index_of_char('.', 0);
+		int second = hostname.index_of_char('.', first+1);
+		if(second != -1 && first != second)
+			hostname = hostname.substring(first+1);
+		string siteURL = uri.get_scheme() + "://" + hostname;
+
+		// download html and parse to find location of favicon
+		var message_html = new Soup.Message("GET", siteURL);
+		if(Settings.tweaks().get_boolean("do-not-track"))
+			message_html.request_headers.append("DNT", "1");
```
jangernert: Yes, I guess so

src/Utils.vala
```diff
@@ -427,45 +438,189 @@ public class FeedReader.Utils : GLib.Object {
...
-			url = url.replace("http://", "");
-		else if(feed_url.has_prefix("https://"))
-			url = url.replace("https://", "");
+		var uri = new Soup.URI(feed_url);
+		string hostname = uri.get_host();
+		int first = hostname.index_of_char('.', 0);
+		int second = hostname.index_of_char('.', first+1);
+		if(second != -1 && first != second)
+			hostname = hostname.substring(first+1);
+		string siteURL = uri.get_scheme() + "://" + hostname;
```
jangernert: I had troubles with some of the sub-domains. I didn't think about the possibility of different icons. Not sure what to do here.

src/FavIconCache.vala
```diff
@@ -70,15 +100,45 @@ public class FeedReader.FavIconCache : GLib.Object {
...
+							Logger.warning("Error loading favicon " + fileInfo.get_name());
+						}
+					}
+				}
+				else
 				{
-					var pixbuf = new Gdk.Pixbuf.from_file_at_scale(iconDirPath + fileName, 24, 24, true);
-					fileName = fileName.substring(0, fileName.length - ".ico".length);
-					m_map.set(fileName, pixbuf);
+					Logger.warning("Error loading favicon " + fileInfo.get_name());
```
jangernert: Yup, will do

src/FavIconCache.vala
```diff
@@ -47,13 +47,43 @@ public class FeedReader.FavIconCache : GLib.Object {
...
+					}
+					catch(GLib.Error e)
+					{
+						Logger.warning("Error loading favicon " + fileInfo.get_name());
+					}
+				}
+				else
+				{
+					Logger.warning("Error loading favicon " + fileInfo.get_name());
+				}
```
jangernert: No this is just a warning. I had some corrupt files (got them from tt-rss) and I just wanted to know what's going on.
I tried both ways of iterating through the files (had a bug and thought it was related to this code). I personally like this way better.


> With next_file, the gboolean return value signifies "end of iteration or error", which requires allocation of a temporary Error.
In contrast, with this function, a false return from iterate *always* means "error". End of iteration is signaled by out_info or out_child being null.

src/FavIconCache.vala
```diff
@@ -47,13 +47,43 @@ public class FeedReader.FavIconCache : GLib.Object {
...
+						var pixbuf = new Gdk.Pixbuf.from_file(iconDirPath + fileName);
+
+						if(pixbuf.get_height() <= 1 && pixbuf.get_width() <= 1)
+						{
+							Logger.warning(@"$fileName too small");
+							continue;
+						}
+
+						pixbuf = new Gdk.Pixbuf.from_file_at_scale(iconDirPath + fileName, 24, 24, true);
+						fileName = fileName.substring(0, fileName.length - ".ico".length);
```
jangernert: All icons are saved with the extension `.ico` regardless of actual image-format. But yes, check should be changed.

plugins/backend/ttrss/ttrssAPI.vala
```diff
@@ -20,11 +20,26 @@ public class FeedReader.ttrssAPI : GLib.Object {
...
 		m_utils = new ttrssUtils();
+		m_session = new Soup.Session();
+		m_session.user_agent = Constants.USER_AGENT;
+		m_session.ssl_strict = false;
+		m_session.authenticate.connect((msg, auth, retrying) => {
+			if(m_utils.getHtaccessUser() == "")
+			{
+				Logger.error("TTRSS Session: need Authentication");
+			}
+			else
```
jangernert: Will do

plugins/backend/ttrss/ttrssAPI.vala
jangernert: This is not the local icon directory, but the one on the server. Apparently the location of the icons are configurable for tt-rss.

plugins/backend/owncloud/OwncloudNewsAPI.vala
```diff
@@ -28,11 +28,25 @@ public class FeedReader.OwncloudNewsAPI : GLib.Object {
...
         m_utils = new OwncloudNewsUtils();
+        m_session = new Soup.Session();
+        m_session.user_agent = Constants.USER_AGENT;
+        m_session.ssl_strict = false;
+        m_session.authenticate.connect((msg, auth, retrying) => {
+			if(m_utils.getHtaccessUser() == "")
+			{
+				Logger.error("ownCloud Session: need Authentication");
+			}
+			else
```
jangernert: Will do

plugins/backend/fresh/freshConnection.vala
```diff
@@ -25,6 +25,16 @@ public class FeedReader.freshConnection {
 		m_settingsTweaks = new GLib.Settings("org.gnome.feedreader.tweaks");
 		m_session = new Soup.Session();
 		m_session.user_agent = Constants.USER_AGENT;
+		m_session.authenticate.connect((msg, auth, retrying) => {
+			if(m_utils.getHtaccessUser() == "")
+			{
+				Logger.error("fresh Session: need Authentication");
+			}
+			else
```
jangernert: Will do

plugins/backend/bazqux/bazquxAPI.vala
```diff
@@ -116,9 +116,9 @@ public class FeedReader.bazquxAPI : GLib.Object {
 			string feedID = object.get_string_member("id");
 			string url = object.has_member("htmlUrl") ? object.get_string_member("htmlUrl") : object.get_string_member("url");
 			string icon_url = object.has_member("iconUrl") ? object.get_string_member("iconUrl") : "";
-			if(icon_url != "" && !m_utils.downloadIcon(feedID, "https:"+icon_url))
+			if(icon_url != "" && !Utils.downloadIcon(feedID, "https:"+icon_url))
```
jangernert: I'm not the original author of this back-end. I just fixed up the code before merging. But I think the "iconUrl" field is actually missing the "https:"

brendanlong: Ah ok.

plugins/backend/ttrss/ttrssAPI.vala
brendanlong: Ah that makes sense.

src/FavIconCache.vala
brendanlong: That makes sense. Since this is a `while(true)` loop, do we also need a `break` here?

src/Utils.vala
brendanlong: Probably not worth the effort right now, but worth thinking about in the future I guess.

src/Utils.vala
brendanlong: Oh wait.. this is a request header, not part of the session. Hm, maybe not.

src/Utils.vala
brendanlong: So maybe order of precedence should be: `rel="icon"`, `rel="shortcut icon"`, `rel="apple-touch-icon"`. Not that it will make much difference in most cases :D

src/Utils.vala
```diff
@@ -427,45 +438,189 @@ public class FeedReader.Utils : GLib.Object {
...
+	{
+		try
+		{
+			string? contents = null;
+			FileUtils.get_contents(filename, out contents);
+			return contents;
+		}
+		catch(Error e)
+		{
+			Logger.warning(@"Utils.checksumForFile: could not read file $filename");
```
brendanlong: This log message refers to `checksumForFile` but the current file is `getFileContent`

src/Utils.vala
```diff
@@ -427,45 +438,189 @@ public class FeedReader.Utils : GLib.Object {
...
+		if(Settings.tweaks().get_boolean("do-not-track"))
+			message_dlIcon.request_headers.append("DNT", "1");
+
+		var status = getSession().send_message(message_dlIcon);
+		if(status == 200)
+		{
+			if(FileUtils.test(local_filename, GLib.FileTest.EXISTS))
+			{
+				var remoteMd5 = GLib.Checksum.compute_for_string(GLib.ChecksumType.MD5, (string)message_dlIcon.response_body.flatten().data);
+				var localMd5 = GLib.Checksum.compute_for_string(GLib.ChecksumType.MD5, getFileContent(local_filename));
```
brendanlong: If we load the entire file every file, it's probably more efficient to just do a string comparison rather than doing an MD5 of both files.

src/Utils.vala
brendanlong: That's probably fine then. I doubt it makes a huge difference in terms of SSD lifetime but I don't really have strong feelings either way.

src/FavIconCache.vala
jangernert: No, this just means a file could not be read, just move on to the next file. We only need to break once the fileInfo is null -> end of the file list.

src/Utils.vala
jangernert: Somehow didn't work. I added the ETag field to the headers of the download request. But I always received a status of 200. Maybe I overlooked something, it's getting very late. I will merge the code anyway. If you want to look into this feel free to do so :)

PR #452, opened 2017-04-17
Use article title as header title

manuq:
From GTK+ docs: "The title should help a user identify the current
view. A good title should not include the application name." [1]

It is already the case for the fullscreen header. This change is for
the column view header.  See other gnome applications for reference,
like Terminal, Polari.

[1] https://developer.gnome.org/gtk3/stable/GtkHeaderBar.html#gtk-header-bar-set-title


jangernert:
Looks good.

PR #458, opened 2017-04-22
Favicon caching improvements

brendanlong:
This makes two major changes:

 - Sends caching headers instead of sending a HEAD request and checking the results.
 - Also sends the last modified time, which a lot of CDN's use instead of ETags.

I changed the format of the icon .txt file to be an INI file with multiple values instead of just etags, but I left the etag-only version as a backup so we don't waste users' bandwidth the first time they run this.


jangernert:
Nice changes. But I don't think the backward compatibility is needed. Etags were only saved since a few days in master. And redownloading the favicons isn't even too painful. So IMO no need to keep it.

PR #459, opened 2017-04-22
Add CircleCI build info

brendanlong:
This will enable CirclECI builds and show a build badge in the README.

You will need to create a CircleCI account and add the project, since it won't let me add a repo I don't own.

I also changed the cmake project type to "C" so it won't fail if we don't have a C++ compiler.

I used Fedora as the build image since that's what I use (easiest for me to setup). Feel free to change it to something else if you prefer. We can make the build *a lot* faster if we make a custom docker image, but I didn't feel like doing that right now.

Also this mainly just tests that everything compiles. We should add some tests at some point.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-project (FeedReader)
+project (FeedReader C)
 cmake_minimum_required (VERSION 2.6)
 list (APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 include_directories(${CMAKE_SOURCE_DIR}/libVilistextum)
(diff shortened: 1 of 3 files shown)


jangernert:
I never used CircleCI before, so I'm a bit lost in their settings. As far as I see they only offer to build on Ubuntu 12.04 or 14.04, which are both too old to build on.

bil-elmoussaoui:
@jangernert The builds run on docker (see CircleCI's yml file) and on the latest fedora container :)

brendanlong:
@jangernert All of the configuration is in the circle.yml so you shouldn't need to configure anything besides adding the project to circleci. The reason the image thing is weird is that I'm using a version 2 CircleCI configuration, which allows you to use any Docker container instead of those two specific Ubuntu images (which would take forever to get the right dependencies). Version 2 is "in beta", so some of the docs don't refer to it yet. https://circleci.com/docs/2.0/

jangernert:
@brendanlong do I have to "Join The Beta" then?

brendanlong:
@jangernert No, I think it's already working: https://circleci.com/gh/jangernert/FeedReader

PR #467, opened 2017-04-23
Favicons

jangernert:
This rips out all the downloading of favicons from the backends and moves it to one single place in the sync()-function. Backends now only need to provide an url where the icon can be found or null if no such url is available.
This should make it easier to parallel the downloading of icons and at some point make the sync-process cancelable.

This branch also drops the `has_icon` column of the feeds table and bumps the db-scheme version to 5. The column was still around from the very early days where FeedReader was tt-rss exclusive.

Lastly the `last_modified` date is compared against current local time to see if last download was more recent than `REDOWNLOAD_FAVICONS_AFTER_DAYS` in which case a re-download is not even attempted.

### DIFF
--- a/plugins/backend/bazqux/bazquxAPI.vala
+++ b/plugins/backend/bazqux/bazquxAPI.vala
@@ -115,7 +115,7 @@ public class FeedReader.bazquxAPI : GLib.Object {
 
 			string feedID = object.get_string_member("id");
 			string url = object.has_member("htmlUrl") ? object.get_string_member("htmlUrl") : object.get_string_member("url");
-			string icon_url = object.has_member("iconUrl") ? object.get_string_member("iconUrl") : "";
+			string? icon_url = object.has_member("iconUrl") ? object.get_string_member("iconUrl") : null;
 
 			string title = "No Title";
 			if(object.has_member("title"))
--- a/plugins/backend/bazqux/bazquxAPI.vala
+++ b/plugins/backend/bazqux/bazquxAPI.vala
@@ -116,10 +116,6 @@ public class FeedReader.bazquxAPI : GLib.Object {
 			string feedID = object.get_string_member("id");
 			string url = object.has_member("htmlUrl") ? object.get_string_member("htmlUrl") : object.get_string_member("url");
 			string icon_url = object.has_member("iconUrl") ? object.get_string_member("iconUrl") : "";
-			if(icon_url != "" && !Utils.downloadIcon(feedID, "https:"+icon_url))
-				icon_url = "";
-			else if(!Utils.downloadFavIcon(feedID, url))
-				icon_url = "something";
 
 			string title = "No Title";
 			if(object.has_member("title"))
@@ -143,9 +139,9 @@ public class FeedReader.bazquxAPI : GLib.Object {
 						feedID,
 						title,
 						url,
-						(icon_url == "") ? false : true,
 						0,
-						categories
+						categories,
+						icon_url
 					)
 			);
 		}
--- a/plugins/backend/ttrss/ttrssAPI.vala
+++ b/plugins/backend/ttrss/ttrssAPI.vala
@@ -212,22 +212,16 @@ public class FeedReader.ttrssAPI : GLib.Object {
 					{
 						var feed_node = response.get_object_element(i);
 						string feed_id = feed_node.get_int_member("id").to_string();
-
-						bool hasIcon = feed_node.get_boolean_member("has_icon");
-
-						if(hasIcon)
-							Utils.downloadIcon(feed_id, m_iconDir + feed_id + ".ico");
-						else
-							hasIcon = Utils.downloadFavIcon(feed_id, feed_node.get_string_member("feed_url"));
+						string? icon_url = feed_node.get_boolean_member("has_icon") ? m_iconDir+feed_id+".ico" : null;
 
 						feeds.add(
 							new feed (
 									feed_id,
 									feed_node.get_string_member("title"),
 									feed_node.get_string_member("feed_url"),
-									hasIcon,
 									(int)feed_node.get_int_member("unread"),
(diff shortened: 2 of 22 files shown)


plugins/backend/bazqux/bazquxAPI.vala
```diff
@@ -116,10 +116,6 @@ public class FeedReader.bazquxAPI : GLib.Object {
 			string feedID = object.get_string_member("id");
 			string url = object.has_member("htmlUrl") ? object.get_string_member("htmlUrl") : object.get_string_member("url");
 			string icon_url = object.has_member("iconUrl") ? object.get_string_member("iconUrl") : "";
```
brendanlong: The empty value should probably be `null`, not `""`

plugins/backend/ttrss/ttrssAPI.vala
```diff
@@ -212,22 +212,16 @@ public class FeedReader.ttrssAPI : GLib.Object {
...
 						var feed_node = response.get_object_element(i);
 						string feed_id = feed_node.get_int_member("id").to_string();
-
-						bool hasIcon = feed_node.get_boolean_member("has_icon");
-
-						if(hasIcon)
-							Utils.downloadIcon(feed_id, m_iconDir + feed_id + ".ico");
-						else
-							hasIcon = Utils.downloadFavIcon(feed_id, feed_node.get_string_member("feed_url"));
+						string? icon_url = feed_node.get_boolean_member("has_icon") ? m_iconDir+feed_id+".ico" : null;
```
brendanlong: Is this going to work, since it's pointing at a local file? I don't know how TTRSS works

plugins/backend/ttrss/ttrssAPI.vala
```diff
@@ -259,21 +253,16 @@ public class FeedReader.ttrssAPI : GLib.Object {
 			{
 				var feed_node = response.get_object_element(i);
 				string feed_id = feed_node.get_int_member("id").to_string();
-				bool hasIcon = feed_node.get_boolean_member("has_icon");
-
-				if(feed_node.get_boolean_member("has_icon"))
-					Utils.downloadIcon(feed_id, m_iconDir + feed_id + ".ico");
-				else
-					hasIcon = Utils.downloadFavIcon(feed_id, feed_node.get_string_member("feed_url"));
+				string? icon_url = feed_node.get_boolean_member("has_icon") ? m_iconDir+feed_id+".ico" : null;
```
brendanlong: Same question here.

src/Structs.vala
```diff
@@ -20,4 +20,75 @@ namespace FeedReader {
...
+
+        public GLib.DateTime? lastModifiedDateTime()
+        {
+            if(last_modified == null)
+                return null;
+
+            GLib.Time time = GLib.Time();
+            time.strptime(last_modified, "%a, %d %b %Y %H:%M:%S %Z");
+
+            return new GLib.DateTime.local(1900 + time.year, 1 + time.month, time.day, time.hour, time.minute, time.second);
```
brendanlong: It seems weird that `time.year` is based at 1900, but I'm not really familiar with GLib.Time().

Here's an example of the spec for this, which says it should always end with the string `GMT`: https://tools.ietf.org/html/rfc7232#section-2.2

(Note that a bunch of people don't actually follow the spec, which is why this class stores it as a string).

src/Utils.vala
```diff
@@ -581,15 +543,28 @@ public class FeedReader.Utils : GLib.Object {
...
+			if(lastMod != null)
+			{
+				var now = new DateTime.now_local();
+				var difference = now.difference(lastMod);
+
+				// icon was already downloaded a few days ago, don't even check if it was updated
+				if((difference/GLib.TimeSpan.DAY) < Constants.REDOWNLOAD_FAVICONS_AFTER_DAYS)
+					return true;
+			}
+
```
brendanlong: You might also want to look at the Expires and Cache-Control headers and save the favicon expiration time, but use 7 days as the minimum: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires

plugins/backend/bazqux/bazquxAPI.vala
jangernert: my bad

plugins/backend/ttrss/ttrssAPI.vala
jangernert: like I mentioned in the last pull-request,m_iconDir is the path to the icons on the server
> icons_url - path to icons when requesting them over http

src/Structs.vala
jangernert: > Years since 1900

https://valadoc.org/glib-2.0/GLib.Time.year.html

> Months since January – [0, 11]

https://valadoc.org/glib-2.0/GLib.Time.month.html

src/Utils.vala
jangernert: Okay, maaaaayyybe I'll do that at some point. I'll add it to the bottom of the todo-list :D

plugins/backend/ttrss/ttrssAPI.vala
brendanlong: But don't we try to download that URL with the favicon downloader? Since it's not on our local machine, I'm confused about how that could work.

jangernert: m_iconDir is the base url to the ttrss-instance (in my case `http://nas/tt-rss/`) plus the `icons_url` from the API. So together something like:
`http://nas/tt-rss/path/to/icons/`
To download the icon for the feed with ID 23 we just add `23.ico`.
The complete url for feed 23 would therefore be
`http://nas/tt-rss/path/to/icons/23.ico`

brendanlong: Ohhh ok 👍

PR #468, opened 2017-04-23
refactor the progress bar

### LINKED ISSUE #427: Load bar instead of a circle
Currently, we see a loading circle appearing in the middle of an article when it's loaded. Would you like to implement this as a 1/2 px loading bar on the very top of the page instead? Then it would be outside of the article, and I've seen this implementation in other Gnome applications, too. :)

bil-elmoussaoui:
Not ready yet ! Closes #427


jangernert:
You broke the overlay-chain :D 
I made some other minor changes. The progress bar is now not inside a box but a overlay. This way the rest of the content is not jumping up and down as the progress-bar is shown and hidden again.

Issue #240, opened 2016-10-10
Too small article view


fbruetting:
I don't know if this is your scope, but with a very small amount of text, the image gets cut off.

![auswahl_009](https://cloud.githubusercontent.com/assets/10520767/19251847/c4117fa6-8f41-11e6-916b-3bee2a403321.png)

jangernert:
Maybe we should ask @bil-elmoussaoui.

bil-elmoussaoui:
Can you give me the link of the article? or at least the html part of code so i can run some tests

fb87:
http://www.golem.de/news/google-pixel-einiges-ist-fuer-telekom-kunden-anders-1610-123700.html

jangernert:
Do you have the content-grabber deactivated? :P

fb87:
No, I just set the dconf switch to kill the daemon when teh application is closed.

bil-elmoussaoui:
I will investigate this and push a fix soon :+1:

fb87:
This is another bug I encountered in master, which may have the same problem root:
![auswahl_001](https://cloud.githubusercontent.com/assets/10520767/20114527/f137334e-a5f4-11e6-917c-26007aa5aeb1.png)

It's due to this article:
https://www.heise.de/newsticker/meldung/Finnland-DDoS-Attacke-auf-Heizungssteuerung-3459730.html?wt_mc=rss.ho.beitrag.rdf

bil-elmoussaoui:
@fb87 sorry for the delay. I will push a fix tomorrow evening! Thanks

bil-elmoussaoui:
I couldn't get that article, can you somehow send me the html file of the page? just click on ctrl +S on your browser and either put the content of it here or email me at bil[dot]elmoussaoui[at]gmail[dot]com?

fb87:
Hi, the first glitch got corrected in the meantime while the second issue still persists – nevertheless, here are both pages. Link works as soon as upload finishes.

https://dl.dropboxusercontent.com/u/327667/FR_bugs_pages.tar.xz

bil-elmoussaoui:
@fb87 I couldn't find anything related to your screenshot on your html file. Use a SQLite manager to read your feedreader database (you will find it in `.local/share/feedreader/data/*.db` and look for the article using the search functionality. This will help you look for the article using the search by title of article, once it's done , copy the content of article and paste it here on github; 
I need the exact tags used here in order to fix it..

[...]

PR #479, opened 2017-05-13
start working on themes

### LINKED ISSUE #299: Better article theming
In order to get different themes to work with FeedReader, I was thinking of a better idea. As the current article theming uses the same style.css and index.html, it's better to add each theme on a different folder under ArticleView.
For example: 
ArticlesView/InstaPaper/
ArticleView/Midnight/
And the UI will show the folder names as the name of the theme, and the index.html file in the currently used theme will be used. 
I will implement this tomorrow if you're okay with that @jangernert ? And I will also add a new beautiful theme :) 
The reason why this should be implemented is to provide the possibility to modify the position of the date and time, the article title, the article content and all of that stuff however the theme creator would like to do.

bil-elmoussaoui:
Should fix #299 #240  once it's ready


src/ArticleTheme.vala
```diff
@@ -0,0 +1,87 @@
+using Gee;
+
+public class  FeedReader.ArticleTheme {
+    static Array<HashMap> themes = new Array<HashMap> ();
+
+    public FeedReader.ArticleTheme() {
+      // Local themes
+      string theme_dir = GLib.Environment.get_home_dir() + "/.feedreader/themes/";
```
jangernert: I'm pretty sure you can't use this path with flatpaks. Also there is already `$HOME/.local/feedreader`.
You can access that via `GLib.Environment.get_user_data_dir() + "/feedreader/"`

bil-elmoussaoui: I will fix that :)

bil-elmoussaoui:
@jangernert Everything should work fine, except a small issue that i couldn't figure out (little tired :/)
The installed themes are correctly detected on both locations global and local one, the fallback theme (saved on gresource works fine too), but the length of the static variable `themes` on `ArticleTheme` is always empty (i've added some printing on my local machine and the themes are correctly detected.. but not added to the static Variable)
My Vala skills aren't that good as yours, can you take a look? 
Thanks!

src/ArticleTheme.vala
```diff
@@ -0,0 +1,109 @@
...
+                    case "author":
+                      author = (string) val;
+                    break;
+                    case "name":
+                      theme_name = (string) val;
+                    break;
+                  }
+                }
+              }catch(Error err){
+                corrupted_theme = true;
```
brendanlong: It would be good to log the error here for debugging purposes.

src/ArticleTheme.vala
```diff
@@ -0,0 +1,109 @@
...
+      bool corrupted_theme = true;
+      string theme_name = "";
+      string author = "";
+      try {
+        Dir theme_dir = Dir.open(theme_path, 0);
+        string ? name = null;
+        while ((name = theme_dir.read_name()) != null){
+          if (name == "theme.json"){
+              corrupted_theme = false;
+              string path = Path.build_filename(theme_path, name);
```
brendanlong: It looks like we're reading through the directory but only doing something with one file. Could we replace this `while` loop with:

```
// etc.
string author = "";
try {
  string path = Path.build_filename(theme_path, "theme.json");
  Json.Parser parser = new Json.Parser();
  // etc.
```

src/ArticleTheme.vala
```diff
@@ -0,0 +1,109 @@
...
+                    break;
+                  }
+                }
+              }catch(Error err){
+                corrupted_theme = true;
+              }
+          }
+        }
+      }catch(GLib.FileError err){
+
```
brendanlong: Logging something here would be nice too.

src/ArticleTheme.vala
```diff
@@ -0,0 +1,109 @@
...
+//	but WITHOUT ANY WARRANTY; without even the implied warranty of
+//	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//	GNU General Public License for more details.
+//
+//	You should have received a copy of the GNU General Public License
+//	along with FeedReader.  If not, see <http://www.gnu.org/licenses/>.
+using Gee;
+
+public class  FeedReader.ArticleTheme {
+    private static ArrayList<HashMap<string, string>> ? themes = null;
```
brendanlong: Consider making the ThemeInfo into a struct:

```
public struct ThemeInfo {
  string name;
  string author;
  bool is_corrupted;
  // etc.
}

private static ArrayList<ThemeInfo>? themes = null;
// etc.
```

Since we know the contents of the map, using a struct is more efficient and easier to work with.

You might also consider making this a map from theme name to theme info so we can more easily lookup themes:

```
private static HashMap<string, ThemeInfo>? = null;

public static HashMap<string, ThemeInfo> getThemes() { ... }

public static ThemeInfo? getTheme(string name)
{
  var themes = getThemes();
  if (!themes.has(name))
    return null;
  return themes.get(name);
}

public static bool exists(string name)
{
  return getTheme(name) != null;
}
```

src/ArticleTheme.vala
```diff
@@ -0,0 +1,109 @@
...
+
+    public static void grabThemes(string location) {
+      try{
+          Dir dir = Dir.open(location, 0);
+          string ? name = null;
+          while ((name = dir.read_name()) != null){
+            string path = Path.build_filename(location, name);
+            if(FileUtils.test(path, FileTest.IS_DIR)){
+              var themeInfo = getThemeInfo(path);
+              if (themeInfo.has_key("corrupted") == false){
```
brendanlong: Since this is the only thing we use `"corrupted"` for, consider just returning `null` when the theme is corrupt.

src/ArticleTheme.vala
```diff
@@ -0,0 +1,109 @@
...
+                themes.add(themeInfo);
+              }
+            }
+          }
+      } catch (GLib.FileError err){
+
```
brendanlong: One more case where logging would be nice (or just don't catch the exception).

src/UtilsUI.vala
```diff
@@ -56,29 +56,52 @@ public class FeedReader.UtilsUI : GLib.Object {
...
-		}
-		catch(GLib.Error e)
-		{
-			Logger.error("Utils.buildArticle: %s".printf(e.message));
-		}
+    string template = "";
+    string css = "";
+    try
+    {
+        if (ArticleTheme.isExists(theme) == false) {
```
brendanlong: It would be simpler to do `if (!ArtistTheme.isExists(theme))`

src/ArticleTheme.vala
```diff
@@ -0,0 +1,109 @@
...
+      }
+    }
+
+    public static bool isExists(string theme_location){
```
brendanlong: I would prefer if this was named `exists()`

src/UtilsUI.vala
```diff
@@ -56,29 +56,52 @@ public class FeedReader.UtilsUI : GLib.Object {
...
 
 		string html_id = "$HTML";
 		int html_pos = article.str.index_of(html_id);
 		article.erase(html_pos, html_id.length);
 		article.insert(html_pos, html);
 
+    string css_id = "$CSS";
+    int css_pos = article.str.index_of(css_id);
+    article.erase(css_pos, css_id.length);
+    article.insert(css_pos, css);
```
brendanlong: Would [`string.replace()`](https://valadoc.org/glib-2.0/string.replace.html) work here?

data/ArticleView/default/article.html
```diff
@@ -0,0 +1,33 @@
...
+    <body style="font-size:$FONTSIZEpt;font-family:$FONTFAMILY;">
+
+        <div id="feedreader-container">
+            <article>
+                <header class="unselectable">
+                    <div id="source" style="font-size:$SMALLSIZEpt;">$FEED</div>
+                    <div id="title" target="_blank" style="font-size:$LARGESIZEpt;"><a href="$URL">$TITLE</a></div>
+                    <div id="author" style="font-size:$SMALLSIZEpt;">
+                         $AUTHOR
+                    </div>
```
brendanlong: Would it be make sense to use more semantic tags like `<h2>` instead of `<div>`?

src/ArticleTheme.vala
bil-elmoussaoui: The code here is still not complete, i would like to check if there's an article.html & style.css file too. Otherwise the theme will be correct. But yeah it's better to use this :+1:

src/ArticleTheme.vala
bil-elmoussaoui: Yeah much better!

src/UtilsUI.vala
bil-elmoussaoui: Javascript habits :p

src/UtilsUI.vala
bil-elmoussaoui: It was used before by @jangernert, i will see what can i do. But i was going to remove including the CSS file here as it's not needed. it's even better to just load the html file, this way third-party themes could contain images, fonts that we shouldn't load by ourselves (vala code)

data/ArticleView/default/article.html
bil-elmoussaoui: The theme isn't ready, it's just an example of how things  works :+1:  
Once the Vala code is ready, i will work on two/three themes that will replace the default ones

brendanlong: Oh cool, that makes sense :)

src/UtilsUI.vala
brendanlong: :+1:

bil-elmoussaoui:
@brendanlong Thanks for the review again! I've pushed few modifications on my latest commit :)

src/ArticleTheme.vala
```diff
@@ -14,14 +14,20 @@
...
 
+public struct ThemeInfo {
+    string name;
+    string author;
+    string path;
+}
+
 public class  FeedReader.ArticleTheme {
-    private static ArrayList<HashMap<string, string>> ? themes = null;
+    private static HashMap<string, ThemeInfo?> ? themes = null;
```
brendanlong: Why is the second type argument `ThemeInfo?`? Are there cases where we would want them theme to `null`?

bil-elmoussaoui: Not at all, it's just valac complained about not using themeInfo?

bil-elmoussaoui:
@jangernert Can you see if everything is alright for you? :+1:

jangernert:
Just wanted to mention the cmake files ^^
As `default` we should imo straight up port the current default theme and just strip out all the color variants

bil-elmoussaoui:
@jangernert I will do that :)

jangernert:
Code seems fine and worked flawlessly for me. The only thing missing is to have the 2 themes on the same quality level as the current themes. Otherwise it will be a downgrade from a user point of view.

bil-elmoussaoui:
Yeah, lt's what i will work on those days :)

PR #480, opened 2017-05-13
local plugin: Prefer <content:encoded> in RSS feeds

### LINKED ISSUE #463: Wordpress blog posts are truncated
For some reason it seems any Wordpress blog feed's items get shown truncated (with a "Continue reading blah blah" link appended) instead of showing the full post contents, like Liferea does.

An example of an affected blog: https://feaneron.com/feed/ (the contents are fully visible in the feed's source code)

brendanlong:
<content:encoded> is an extension to RSS which is basically like
<content> in ATOM. Wordpress uses it for the full text content,
so we should prefer it when it exists.

Also fixed a call to nxmle_free() which should just be free().

Fixes #463

### DIFF
--- a/plugins/backend/local/libmrss/mrss_parser.c
+++ b/plugins/backend/local/libmrss/mrss_parser.c
@@ -365,7 +365,7 @@ __mrss_parser_atom_entry (nxml_t * doc, nxml_data_t * cur, mrss_t * data)
 	  {
 	  	if (item->description)
 	  	{
-	  	  nxml_free(item->description);
+	  	  free(item->description);
 	  	  item->description = NULL;
 	  	}
 	    __mrss_parser_atom_string (doc, cur, &item->description,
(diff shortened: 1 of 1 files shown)


plugins/backend/local/libmrss/mrss_parser.c
```diff
@@ -365,7 +365,7 @@ __mrss_parser_atom_entry (nxml_t * doc, nxml_data_t * cur, mrss_t * data)
 	  {
 	  	if (item->description)
 	  	{
-	  	  nxml_free(item->description);
+	  	  free(item->description);
```
brendanlong: If you look in nxmle_get_string(), it calls nxml_get_string(), which creates the string with `strdup()`, so `free` is the right thing to call here, not `nxmle_free`.

jangernert:
awesome work :)

PR #482, opened 2017-05-14
Minor cleanups

brendanlong:
This fixes some build warnings, changes non-error logs to 'info' level, and fixes some typos.


jangernert:
Btw: you can commit small stuff like this directly to master if you want ;)
I don't think anyone would argue about changes like this.

PR #483, opened 2017-05-14
Feedbin improvements

brendanlong:
Various improvements to the Feedbin backend. This makes setup a lot faster (lets us show articles sooner) and fixes some crashes.

Depends on #482

PR #484, opened 2017-05-14
Always download favicons in the background.

brendanlong:
This could still be improved, since we're reading files and things
synchronously, but at least we do one favicon at a time now.

### DIFF
--- a/plugins/backend/local/SuggestedFeedRow.vala
+++ b/plugins/backend/local/SuggestedFeedRow.vala
@@ -62,8 +62,8 @@ public class FeedReader.SuggestedFeedRow : Gtk.ListBoxRow {
 		show_all();
 
 		var uri = new Soup.URI(url);
-		downloadIcon.begin(uri.get_host(), "/tmp/", uri.get_scheme() + "://" + uri.get_host(), (obj, res) => {
-			bool success = downloadIcon.end(res);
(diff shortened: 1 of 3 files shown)


brendanlong:
With this and #483, my massive feed list starting being usable in about 7 seconds, so that's a lot better.

jangernert:
I took a quick look and seems fine :)
I'll take a closer look tomorrow and try to understand everything you changed.

jangernert:
Only question I have:
How is this currently handled in the UI? Since all the favicons are downloaded in the background the feed list will be displayed with all feeds having the default rss-icon, right?
And as far as I remember `updateFeedList()` only refreshes the unread counters at the moment.
So it would probably be nice to send a signal to the UI after `Utils.getFavIcons` so all favicons could be checked again.
I'll get to it tonight.

brendanlong:
The icons in the article view don't refresh until something causes the article list to change (like clicking on a different list). I didn't think to expand the feed list, but I'd guess it doesn't refresh itself until restart(?), which would be a bigger problem.

jangernert:
Don't worry, I'll take care of this.

jangernert:
818f7cb393956e1fac17e5b0680d7de5f6b36740
Haven't actually tested it with logging in to a new account, but judging from the logs on my account with all favicons already downloaded everything is working as it should.

brendanlong:
Looks good to me. At some point we could probably make it nicer by reloading individual favicons as we download them, but for normal people this will probably complete relatively quickly anyway.

PR #486, opened 2017-05-20
display possible error after adding a feed

### LINKED ISSUE #461: Autodetect protocols when adding local feeds
When I try to add a feed, my way of doing it will typically be to just tell the app to add "example.com/blog". Yet FeedReader silently refuses to add it unless I add https:// or http:// in front... which is unnecessary. Users don't care about protocols. Just automatically try https, then http, then whatever else is supported. And if somehow there was no feed found, show me an error message to let me know why.

jangernert:
This enables the back-ends to specify what went wrong when trying to add a feed and display the error in the UI as a small popover on the '+'-button.
The popover could use some styling. Some of the back-ends could use some more detailed error messages but at least for tt-rss and the local back-end everything seems to work fine.

First step to fix #461 
edit: theoretically similar changes could be made to actions like removing, renaming and moving feeds and categories. But since adding a feed is the most likely action to fail this has to do for now.

### DIFF
--- a/plugins/backend/bazqux/bazquxAPI.vala
+++ b/plugins/backend/bazqux/bazquxAPI.vala
@@ -446,7 +446,7 @@ public class FeedReader.bazquxAPI : GLib.Object {
 		m_connection.send_post_request("rename-tag", msg.get());
 	}
 
-	public void editSubscription(bazquxSubscriptionAction action, string feedID, string? title = null, string? add = null, string? remove = null)
+	public bool editSubscription(bazquxSubscriptionAction action, string feedID, string? title = null, string? add = null, string? remove = null)
 	{
 		var msg = new bazquxMessage();
 		msg.add("output", "json");
@@ -476,6 +476,11 @@ public class FeedReader.bazquxAPI : GLib.Object {
 			msg.add("r", remove);
 
 
-		m_connection.send_post_request("subscription/edit", msg.get());
+		var response = m_connection.send_post_request("subscription/edit", msg.get());
+
+		if(response.status == 200)
+			return true;
+
+		return false;
 	}
 }
--- a/plugins/backend/bazqux/bazquxAPI.vala
+++ b/plugins/backend/bazqux/bazquxAPI.vala
@@ -478,9 +478,6 @@ public class FeedReader.bazquxAPI : GLib.Object {
 
 		var response = m_connection.send_post_request("subscription/edit", msg.get());
 
-		if(response.status == 200)
-			return true;
-
-		return false;
+		return response.status == 200;
 	}
 }
--- a/plugins/backend/bazqux/bazquxInterface.vala
+++ b/plugins/backend/bazqux/bazquxInterface.vala
@@ -175,18 +175,26 @@ public class FeedReader.bazquxInterface : Peas.ExtensionBase, FeedServerInterfac
 		return m_api.ping();
 	}
 
-	public string addFeed(string feedURL, string? catID, string? newCatName)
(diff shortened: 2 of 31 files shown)


plugins/backend/bazqux/bazquxAPI.vala
```diff
@@ -476,6 +476,11 @@ public class FeedReader.bazquxAPI : GLib.Object {
 			msg.add("r", remove);
 
 
-		m_connection.send_post_request("subscription/edit", msg.get());
+		var response = m_connection.send_post_request("subscription/edit", msg.get());
+
+		if(response.status == 200)
+			return true;
+
+		return false;
```
brendanlong: You could just `return response.status == 200` here.

plugins/backend/bazqux/bazquxInterface.vala
```diff
@@ -175,18 +175,26 @@ public class FeedReader.bazquxInterface : Peas.ExtensionBase, FeedServerInterfac
...
 		}
 		else
 		{
-			m_api.editSubscription(bazquxAPI.bazquxSubscriptionAction.SUBSCRIBE, "feed/"+feedURL, null, catID);
+			success = m_api.editSubscription(bazquxAPI.bazquxSubscriptionAction.SUBSCRIBE, "feed/"+feedURL, null, catID);
 		}
-		return "feed/" + feedURL;
+
+		if(!success)
+			errmsg = @"bazqux could not subscribe to $feedURL";
```
brendanlong: I wonder if it would be simpler to throw an exception here than to use a bool return + multiple out parameters.

Also it would be nice if we could be more specific in the error message (failure due to status x, failure because catID and netCatName are both null).

Longer term it would be nice to split this into two functions, one of which takes `string catID` and the other which takes `string newCatName` (both non-null).

plugins/backend/oldreader/oldreaderInterface.vala
```diff
@@ -176,18 +176,26 @@ public class FeedReader.OldReaderInterface : Peas.ExtensionBase, FeedServerInter
...
 		}
 		else
 		{
-			m_api.editSubscription(OldReaderAPI.OldreaderSubscriptionAction.SUBSCRIBE, {"feed/"+feedURL}, null, catID);
+			success = m_api.editSubscription(OldReaderAPI.OldreaderSubscriptionAction.SUBSCRIBE, {"feed/"+feedURL}, null, catID);
 		}
-		return "feed/" + feedURL;
+
+		if(!success)
+			errmsg = "The old reader could not add %s";
```
brendanlong: This `%s` seems to be missing an `sprintf`

plugins/backend/owncloud/OwncloudNewsAPI.vala
```diff
@@ -435,15 +435,31 @@ public class FeedReader.OwncloudNewsAPI : GLib.Object {
...
+		feedID = 0;
+
+		switch(message.getStatusCode())
+		{
+			case 409:
+				errmsg = "Feed already added";
+				return true;
+			case 422:
+				errmsg = "ownCloud can't read the feed";
+				break;
```
brendanlong: It would be nice to include the error code in errmsg.

plugins/backend/bazqux/bazquxInterface.vala
jangernert: I'm not sure if dbus methods can throw exceptions. I haven't found any code using them.
Jup the messages could be more detailed for some of the backends. Maybe I should do a wiki page with all sorts of junior jobs like this.

PR #490, opened 2017-05-24
ArticleViewHeader: improve star and circle buttons

manuq:
- Colorize the shapes when the buttons are hovered or active.  This way
  they are consistent with the same icons in the article list sidebar.

- Replace the custom widget with Gtk.ToggleButton styled with CSS.

- Add tooltips for those buttons.  The buttons can now have a tooltip,
  which has been said to be buggy in the custom HoverButton widget.  A
  tooltip is useful for users with disabilities as well as a description
  for unobvious icons, like the circle for Mark unread.


src/Widgets/ArticleViewHeader.vala
```diff
@@ -30,6 +30,15 @@ public class FeedReader.ArticleViewHeader : Gtk.HeaderBar {
 	public signal void popClosed();
 	public signal void popOpened();
 
+	private void onMarkButton()
```
bil-elmoussaoui: Why those are needed ? They do nothing other than calling an other method

manuq:
For more information see my 1st attempt at adding tooltips: https://github.com/jangernert/FeedReader/pull/448 . This change goes further replacing the HoverButton and adding consistency with the same buttons in the sidebar.

src/Widgets/ArticleViewHeader.vala
manuq: I wrapped the signal emition to be able to connect / disconnect the wrapped function. If I try to connect to toggledMarked directly, I get this when compiling:
```
/run/build/feedreader/src/Widgets/ArticleViewHeader.vala:51.33-51.45: error: Argument 1: Cannot convert from `FeedReader.ArticleViewHeader.toggledMarked' to `Gtk.ToggleButton.toggled'
		m_mark_button.toggled.connect(toggledMarked);
		                              ^^^^^^^^^^^^^
```
The connect/disconnect has to happen when the widget is updated from the outside, like from the sidebar for example.

jangernert:
Not a fan of any of those changes.

manuq:
> Not a fan of any of those changes.

Reason? It's OK if you want to reject this without a reason, you are the maintainer and nobody asked for this. I just saw an opportunity to improve the application for something that bothered me as a user. In https://github.com/jangernert/FeedReader/pull/448 you said you'll be happy to merge an improved version, so I discussed it in #gnome-design IRC channel and I gave this a second round.

jangernert:
The comment about the merge was about a fix for tooltips messing up the enter/leave-events, so tooltips could be enabled again with the existing buttons.
The headerbar should either be all colored icons or all symbolic, no mixture of both. It seems more people are in favor of symbolic icons.
Enabling tooltips is fine, but like I said causes troubles with events with the current implementation. I hope this is something that can be fixed or worked around with the big switch to Gtk+ 4. I remember reading something about lots of changes in this area.
For keeping the hoverbutton I don't have a valid objective argument. I just really like preview of the state change on hover.
I hope this does not discourage you from suggesting similar small changes like this.

manuq:
OK @jangernert your explanation sounds reasonable to me. I don'r see a problem doing colored symbolic icons for cases like a star icon or the hearth icon in Corebird (which was my reference both for design and implementation) But I also understand that some people would like all BW symbolic icons for consistency, and that means keeping the custom widget in current GTK+.

PR #492, opened 2017-05-24
Add a .editorconfig file

bil-elmoussaoui:
Editorconfig is supported by almost every text editor/IDE out there (even Builder)

It's used to avoid having spaces/tabs, or different indentation size
If there's any modification to do, just let me know :)

This helps a lot if you switch text editor, or something else :D

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


brendanlong:
:+1:

PR #493, opened 2017-05-25
Use git submodule for grabberconfig

bil-elmoussaoui:
Instead of copying the updates done by the guys at five-filters, it's better to use a git submodule   
@jscurtu is there any modification that should be done on the flatpak package?


jangernert:
Is it possible to apply patches to the submodule?
We still carry a patched config for phoronix to remove double `<div class='content'>`

bil-elmoussaoui:
The patches should be applied to five-filters repository instead. Only if we wish to keep maintaining that huge database by ourselves...

jangernert:
The changes were removed from filefilters. So I don't think it's something they'll merge.
Alternatively we have to rename the `content`-class in `article.html` and `style.css` to `feedreader_content` or something

bil-elmoussaoui:
Yeah, i will make sure of that before we merge #479

bil-elmoussaoui:
We can add a prefix for each class/id like FeedReaderArticleView to be sure that we won't mess with any other content

PR #498, opened 2017-06-04
FeedReader is also packaged in openSUSE

Mailaender:
https://software.opensuse.org/package/feedreader

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,11 @@ sudo dnf install feedreader
 sudo eopkg install feedreader
 ```
 
+### openSUSE : <br/>
+```bash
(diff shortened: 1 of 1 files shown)


bil-elmoussaoui:
Is there any specific reason why you have used an html tag instead of just a new line?
Can you also update the gh-pages branch

Mailaender:
> Is there any specific reason why you have used an html tag instead of just a new line?

I just copied it from the lines above.

bil-elmoussaoui:
Ah sorry didn't saw the whole file (reviewing using mobile...)

PR #500, opened 2017-06-14
it.po updated

luke3v:


bil-elmoussaoui:
Hey, Thanks for the PR. But the translations updates must be done on Transifex, i update the translations before each release

PR #533, opened 2017-08-25
add copy feed url to feed right click menu

### LINKED ISSUE #496: Make it easy to copy feed URLs
Would be nice to have a way to copy the feed's URL, or open the attached website when right-clicking on the feed's list entry.

bil-elmoussaoui:
Fixes #496


jangernert:
instead of adding a new function to dbBase I would rather use `public article? read_article(string articleID)` and read the url from the returned object ;)

bil-elmoussaoui:
It's about getting the FeedURL and not the articleURL

jangernert:
Sorry, `public feed? read_feed(string feedID)` should be the right one. Then use  `feed.getXmlUrl()`

PR #549, opened 2017-09-25
Fix Tiny Tiny RSS plugin

JKingweb:
The Tiny Tiny RSS plugin has been broken for a few months (since at least [May 4, 2017](https://git.tt-rss.org/git/tt-rss/commit/7c9b5a3fe40a641af1cb6003b03b2352576a0c4b)). This patch fixes it.

### DIFF
--- a/data/tt-rss-feedreader-plugin/api_feedreader/init.php
+++ b/data/tt-rss-feedreader-plugin/api_feedreader/init.php
@@ -37,7 +37,7 @@ class Api_feedreader extends Plugin {
 		$label_id = (int)db_escape_string($_REQUEST["label_id"]);
 		if($label_id != "")
 		{
-			label_remove(feed_to_label_id($label_id), $_SESSION["uid"]);
+			Labels::remove(Labels::feed_to_label_id($label_id), $_SESSION["uid"]);
 			return array(API::STATUS_OK);
 		}
 		else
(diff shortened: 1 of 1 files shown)


jangernert:
Great work. Thank you!
Was the API version also bumped with the changes?

JKingweb:
I don't know; I'd have to find where the API lever is defined. I'll check when I have some time, but I can tell you that the level is at 14 in my month-old installation. I should probably update it and make sure my fixes still work against the current commit. 

I noticed the breakage, by the way, because I'm working on implementing the API in [my own RSS server](https://code.mensbeam.com/MensBeam/arsse), and I'm keen on supporting FeedReader natively, with automated testing so things won't break.

JKingweb:
I checked:

1. This patch still works with TT-RSS' head commit
2. The version (defined in `include/version.php`) was 17.4 before the changes to the label functions, and remains so today
3. The API level (defined in `classes/api.php`) was 14 before and remains so today.

jangernert:
Hm, I wonder if the API versions jumped so quickly recently, or if I got the wrong number when I wrote the plugin :D 
Anyway, let's merge this.

jangernert:
I saw your RSS server supports the nextcloud API. Shouldn't that work with FeedReader as well?

JKingweb:
It should! I'm principally a Windows user, so I've not actually tried FeedReader yet, but other clients for NextCloud News seem to work fine.

PR #553, opened 2017-09-29
Remove extraneous slashes in ownCloud URLs

JKingweb:
For many operations with ownCloud/NextCloud, FeedReader sends requests to URLs with extraneous slashes. NextCloud is forgiving of this, but The Arsse (my server implementation) is not, which leaves FeedReader unable to perform several operations.

```
"POST /index.php/apps/news/api/v1-2//folders HTTP/1.1" 501
"POST /index.php/apps/news/api/v1-2//feeds HTTP/1.1" 501
```

Please note I don't have an environment suitable for compiling FeedReader (and therefore testing this fix), but it seems pretty clear the slashes removed in this commit are the cause of the slash doubling.

### DIFF
--- a/plugins/backend/owncloud/OwncloudNewsAPI.vala
+++ b/plugins/backend/owncloud/OwncloudNewsAPI.vala
@@ -215,7 +215,7 @@ public class FeedReader.OwncloudNewsAPI : GLib.Object {
 
 	public void getNewArticles(Gee.List<article> articles, int lastModified, OwnCloudType type, int id)
 	{
-		var message = new OwnCloudNewsMessage(m_session, m_OwnCloudURL + "/items/updated", m_username, m_password, "GET");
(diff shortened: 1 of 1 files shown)


brendanlong:
Note: We guarantee that `m_OwnCloudURL` ends with `/` in OwnCloudNewsUtils.getURL():

```
	public string getURL()
	{
		string tmp_url = Utils.gsettingReadString(m_settings, "url");
		if(tmp_url != ""){
			if(!tmp_url.has_suffix("/"))
				tmp_url = tmp_url + "/";

			if(!tmp_url.has_suffix("/index.php/apps/news/api/v1-2/"))
				tmp_url = tmp_url + "index.php/apps/news/api/v1-2/";

			if(!tmp_url.has_prefix("http://") && !tmp_url.has_prefix("https://"))
					tmp_url = "https://" + tmp_url;
		}

		Logger.debug("OwnCloud URL: " + tmp_url);

		return tmp_url;
	}
```

So I think this is right.

PR #558, opened 2017-10-03
Change ownCloud ping to valid method

JKingweb:
The previous method (a PUT to `api/v1-2/`) is not a part of the API description. Not only does The Arsse reject it, but NextCloud News itself gets confused and redirects to the user's NextCloud Files main page instead. It works, but may not in future NextCloud versions.

Instead, a GET to the `api/v1-2/version` route is both valid and low-overhead, and should continue to work in perpetuity.

I also corrected a repeated typo I noticed while I was looking around.

After this change FeedReader should work flawlessly with an instance of The Arsse.

### DIFF
--- a/plugins/backend/owncloud/OwncloudNewsAPI.vala
+++ b/plugins/backend/owncloud/OwncloudNewsAPI.vala
@@ -550,7 +550,7 @@ public class FeedReader.OwncloudNewsAPI : GLib.Object {
 
 	public bool ping()
 	{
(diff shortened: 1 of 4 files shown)


brendanlong:
Cool, thanks for the pull request! :)

brendanlong:
(API link if anyone else wants it: https://github.com/owncloudarchive/news/wiki/Version-1.2)

PR #560, opened 2017-10-06
Single process

jangernert:
Merge the daemon and ui process into a single process called 'feedreader' (thinking about naming it org.gnome.FeedReader). All the backend plugins are also a single library now.

I'll still look around the codebase and see what can be improved. I think there are still things that can be simplified now.


jangernert:
@brendanlong yeah, we should split them up according to context. But before almost everything was in Utils. Everything that depended on stuff only available in the daemon was in daemonUtils and vice versa.

jangernert:
@brendanlong Okay, IDs are now kept unchanged internally (modifications in the backend are also gone now). When used as file/folder-names IDs will be encoded with base64. There was another issue with one of the database functions accepting semicolon separated lists of IDs. This failed sometimes because IDs of local backend could contain semicolons themselves. This was still code from the time FeedReader was ttrss specific.
I'll merge this now so you can base your feedbin improvements on it.

PR #561, opened 2017-10-07
Feedbin testable

brendanlong:
- Makes the Feedbin API completely standalone so it can easily be linked to other executables (like tests)
 - Makes FeedbinAPI.vala only concerned with Feedbin things (conversion and subscription/feed hacks go in FeedbinInterface.vala)
 - Adds a gtest tests
 - Makes the tests run in CircleCI


brendanlong:
Ok, now the error handling works, but I need to go through and change all of the `int64.parse` calls to check or throw instead of having assert failures.

brendanlong:
@jangernert BTW, it would be relatively easy for me to rebase this on top of your changes so feel free to merge that whenever you're comfortable and I can make this work on top of it.

jangernert:
Sounds nice. I'll have a look at the code later. Just a general question: We want to strip out all the database calls and only reference some essentials like the logger and a subset of utils. Wouldn't this mean we could build the plugins against the vapi of a small library only consisting of said things? This in turn would mean we could write tests against the backend interface and use them for all the backends.

brendanlong:
Yeah that would be really good if we could get the plugins to build on their own :)

PR #565, opened 2017-10-09
Respect GNOME clock format setting when displaying article dates

brendanlong:
I'm not sure if this is the most reasonable way to do it.

Example (notice that the format is the same as my clock):

screenshot from 2017-10-08 21-26-02

### DIFF
--- a/src/Model/Article.vala
+++ b/src/Model/Article.vala
@@ -32,16 +32,23 @@ public class FeedReader.Article : GLib.Object {
 	private int m_lastModified;
 	private int m_pos;
 
-	private static GLib.Settings gnome_settings;
-	private static string clock_format;
+	private static GLib.Settings? gnome_settings;
+	private static bool clock_12_hour = false;
 
 	static construct
 	{
-		gnome_settings = new GLib.Settings("org.gnome.desktop.interface");
-		clock_format = gnome_settings.get_string("clock-format");
-		gnome_settings.changed["clock-format"].connect(() => {
-			clock_format = gnome_settings.get_string("clock-format");
-		});
+		// Lookup the schema in a complicated way so we don't require users
+		// to be running GNOME Shell
+		var schema_source = SettingsSchemaSource.get_default();
+		var schema = schema_source.lookup("org.gnome.desktop.interface", true);
+		if(schema != null)
+		{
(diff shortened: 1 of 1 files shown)


src/Model/Article.vala
```diff
@@ -32,6 +32,9 @@ public class FeedReader.Article : GLib.Object {
 	private int m_lastModified;
 	private int m_pos;
 
+	private static string clock_format =
+		new GLib.Settings("org.gnome.desktop.interface").get_string("clock-format");
```
brendanlong: Is this a reasonable thing to do? I don't really understand Vala's `static` semantics. It works, but I don't know if it's a bad idea.

jangernert: I'm not sure either. I would assume the settings object is created once and since there is no reference holding it, runs out of scope and gets freed.
But luckily it's quite easy to find out. Just take a look at the gerated c code ;)

src/Model/Article.vala
```diff
@@ -157,33 +160,40 @@ public class FeedReader.Article : GLib.Object {
...
 			if(date_day == now_day)
 			{
-				return m_date.format("%H:%M");
+				addTime = true;
 			}
-			else if(date_day == now_day-1)
+			else if(date_day == now_day -1)
 			{
-				return _("Yesterday") + m_date.format(", %H:%M");
+				formats.add("Yesterday");
```
jangernert: Is this still translatable, or am I missing something here?

brendanlong: Good catch, I didn't mean to change that :\

bil-elmoussaoui:
Shouldn't we check if the schema exists and fallback to the default format we had before? Otherwise this will crash the application:/ unless we are going to force users to use GNOME or flatpak

brendanlong:
> Shouldn't we check if the schema exists and fallback to the default format we had before? Otherwise this will crash the application:/ unless we are going to force users to use GNOME or flatpak

Good point. I'll do that too.

jangernert:
Btw: It would be nice to keep member variable names with the `m_` prefix (or just `_` if you prefer that, although `m_` would be better, since it's already used everywhere). Otherwise I will keep wondering why some variable appears all of a sudden, only to realize it is a member a minute later :D

PR #569, opened 2017-10-09
Favicon expiration

brendanlong:
This should make it work better when you add a new feed (don't need to redownload all favicons), and I also fixed a bug in the local feed setup (it was using the old un-encoded feed ID's).

I'm tempted to make an IconWidget that does this stuff in the background on load, but this seemed good enough for now.

### DIFF
--- a/plugins/backend/local/SuggestedFeedRow.vala
+++ b/plugins/backend/local/SuggestedFeedRow.vala
@@ -70,7 +70,7 @@ public class FeedReader.SuggestedFeedRow : Gtk.ListBoxRow {
 
 	private async void load_favicon(Gtk.Stack iconStack, Feed feed, string iconURL)
 	{
-		bool success = yield Utils.downloadIcon(feed, iconURL, null);
+		bool success = yield Utils.downloadFavIcon(feed, iconURL);
 
 		Gtk.Image? icon = null;
 		if(success)
--- a/plugins/backend/local/SuggestedFeedRow.vala
+++ b/plugins/backend/local/SuggestedFeedRow.vala
@@ -63,34 +63,31 @@ public class FeedReader.SuggestedFeedRow : Gtk.ListBoxRow {
 
 		var uri = new Soup.URI(url);
 		var fakeFeed = new Feed(uri.get_host(), "", "", 0);
-		Utils.downloadIcon.begin(fakeFeed, iconURL, null, "/tmp/", (obj, res) => {
-			bool success = Utils.downloadIcon.end(res);
-			Gtk.Image? icon = null;
-
-			if(success)
-			{
-				try
-				{
-					string filename = "/tmp/" + uri.get_host().replace("/", "_").replace(".", "_") + ".ico";
-					Logger.debug("load icon %s".printf(filename));
-					var tmp_icon = new Gdk.Pixbuf.from_file_at_scale(filename, 24, 24, true);
-					icon = new Gtk.Image.from_pixbuf(tmp_icon);
-				}
-				catch(GLib.Error e)
-				{
-					Logger.error("SuggestedFeedRow.constructor: %s".printf(e.message));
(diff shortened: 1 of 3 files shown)


jangernert:
Seems fine :)
From what I understand the metadata with the expires timestamp is written now even when the download fails? 

Idea: We should implement a way to update the favicon of a specific feed in the UI (probably just alter the current signal to have a feed object as an argument). Move the downloading of favicons into the FavIconCache (maybe rename FavIconManager). The FavIconManager would hold the metadata together with the actual icon. Every time an icon is requested and found in the cache, it is returned immediatly while the expiration date is checked async and in case it has expired will be redownloaded. If an icon is not already in the cache the default icon is returned immediatly, the FavIconManager checks for an existing metadata file and downloads the icon in case no metadata was found or the last download attempt was more than REDOWNLOAD_FAVICONS_AFTER_DAYS ago.
After a successful download the FavIconManager sends out a signal for the UI to reload the icon of the specific feed.
Which in turn makes the UI query the FavIconManager again, which can now return the downloaded icon.
This way the UI does not have to care about anything. It just needs to request an icon from the FavIconManager and reload its icon once it is told to.

I hope I could express my idea well enough for you to understand :D 
Maybe this is something similar to what you had in mind with the IconWidget.

jangernert:
Do you think having and managing all these callbacks is worth the possible speed tradeoff? Just emmiting a signal everytime a new icon is downloaded (something that shouldn't happen too often) and reloading it everywhere seems way easier to me.

Timeouts wouldn't be needed imo. If a feed changes its icon, the update would be triggered as soon as a new article of that feed pops up or when the user changes the article list to something that shows an article of that feed. The only case it wouldn't update the favicon is if no new articles of that feed are added to the list and if the user never clicks on the feed or its category to view the articles. That is a pretty fair tradeoff imo.

jangernert:
Merged!
Also, did you remove the `last-favicon-update` setting as it is no longer needed?

brendanlong:
Maybe signals are the right way to go. The things I wanted to avoid are:

 - Making every favicon widget refresh itself every time we load a favicon
 - Having to duplicate the lookup code

I'd prefer this:

```
m_icon = default;
FaviconCache.get_default().subscribe(feed_id, (pixbuf) => {
  m_icon.pixbuf = pixbuf;
});
```

Over this:

```
m_icon = default;
var cache = FaviconCache.get_default();
cache.new_icon.connect((feed_id, pixbuf) => {
  // note: this gets called for every feed_id / icon widget combination
  if(feed_id == m_feed_id)
    m_icon = pixbuf;
});

// in case the icon is already loaded
cache.getIcon.begin(feed_id, (obj, res) => {
  var pixbuf = cache.getIcon.end(res);
  if(pixbuf != null)
    m_icon.pixbuf = pixbuf;
});
```

But maybe there's a better way of writing the second version?

jangernert:
I was more thinking along the lines:
```
construct
{
  m_icon = default;
  cache.getIcon(feed, (obj, res) => {
    var pixbuf = cache.getIcon.end(res);
    if(pixbuf != null)
      ReloadIcon(pixbuf);
  });
}

public void ReloadIcon(Pixbuf p)
{
  m_icon.data = p;
}
```

If a new icon was downloaded or updated the cache could either emmit a signal which is picked up by the mainwindow or call the mainwindow directly to update all icons of that particual feed. The mainwindow would then tell the visible view (currently only columnview available) to update that icon. The view would tell the feedlist and the articlelist to update that icon. Both lists would look through their rows and tell every row that matches that id to reload its icon.
Basically the same way we do it now with the additional feed argument and check against the feedid for each row.

Offtopic:
The iteration over every row in the ui is something I want to change in the future, but probably only around the time we make the change to gtk4 or in preparation for it. FeedReader was my first UI in Gtk and my first UI that was more than a window with a couple of buttons and lists on it :D

brendanlong:
I suspect the callback thing would be worth it if it would let us get rid of the iteration over every row :\ I'll try to figure out how complicated it would be, but I suspect making the favicon cache slightly smarter would let us remove a bunch of icon updating code.

Issue #571, opened 2017-10-12
Load favicons on-demand


brendanlong:
This is writeup of something me and @jangernert were talking about (feel free to edit if you think this should be done differently).

Right now we load favicons at the end of each sync, by looking through every feed and trying to download their favicon (with checks to make sure we don't do it too often), then reloading all favicons.

This isn't ideal:

 - We download icons we're not going to use (for example, if a feed has no unread articles and the user has chosen to only show unread)
    - Arguably, this could be a feature, if a user is reading offline and then decides to look through 'all' articles
 - When we add a new feed, we have to reload all icons instead of just loading the new one
 - It's hard for us to download icons in parallel, since Vala doesn't seem to have a way to run multiple async functions at once and wait for them all (although maybe this exists and I just can't find it)

Proposal:

1. Move favicon downloading object from Utils.vala to FavIconCache.vala
2. Change `FavIconCache.load()` to call `downloadFavIcon()` before trying to load the icon
3. Add a callback `FavIconCache.loaded(string feed_id, Gdk.Pixbuf pixbuf)`
4. Make `FavIconCache.load` trigger `loaded` when a new icon is downloaded
5. Make sure we only try to download each icon once (`Set<string> m_loading_icons`?)
6. Remove all of the code to trigger favicon downloading after a refresh

Things that would be nice but probably aren't required in the initial version:

 - Remove pixbufs from the cache when feeds are deleted (using a `Map<string, WeakRef<Gdk.Pixbuf>>` might work)
 - Only notify subscribers if the icon they want is loaded (have a way to subscribe to "feed 3's icon was loaded" instead of "some feed's icon was loaded")
 - A way to trigger downloading all favicons for offline mode?

jangernert:
I started working on this in the icon_refactor branch. I probably have to entertain the girlfriend for the rest of the weekend. So if you have some time on your hands you could pick it up from there ;)

brendanlong:
Cool! I also have a lot going on this weekend but I'll take a look if I have time

PR #573, opened 2017-10-18
FavIconManager refactor

jangernert:
I think that should be it. But I could have still missed or misunderstood something in #571.

### DIFF
--- a/src/FavIconManager.vala
+++ b/src/FavIconManager.vala
@@ -13,20 +13,20 @@
 //	You should have received a copy of the GNU General Public License
 //	along with FeedReader.  If not, see <http://www.gnu.org/licenses/>.
 
-public class FeedReader.FavIconCache : GLib.Object {
+public class FeedReader.FavIconManager : GLib.Object {
 
 	private Gee.HashMap<string, Gdk.Pixbuf> m_map;
-	private static FavIconCache? m_cache = null;
+	private static FavIconManager? m_cache = null;
 
-	public static FavIconCache get_default()
+	public static FavIconManager get_default()
 	{
 		if(m_cache == null)
-			m_cache = new FavIconCache();
+			m_cache = new FavIconManager();
 
 		return m_cache;
 	}
 
-	private FavIconCache()
+	private FavIconManager()
 	{
 		m_map = new Gee.HashMap<string, Gdk.Pixbuf>();
 	}
@@ -43,7 +43,7 @@ public class FeedReader.FavIconCache : GLib.Object {
 			stream.close();
 			if(pixbuf.get_height() <= 1 && pixbuf.get_width() <= 1)
 			{
-				Logger.warning(@"FavIconCache: $fileName is too small");
+				Logger.warning(@"FavIconManager: $fileName is too small");
 				return;
 			}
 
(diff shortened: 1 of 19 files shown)


src/FavIconManager.vala
```diff
@@ -0,0 +1,321 @@
...
+		catch (IOError.NOT_FOUND e)
+		{
+			//Logger.debug(@"FavIconManager: Icon $fileName does not exist");
+
+			if(!firstTry)
+				return;
+
+			bool success = yield downloadFavIcon(feed);
+			if(success)
+				yield load(feed, false);
```
brendanlong: I think that if an icon doesn't exist, this will cause us to try downloading it every time we display one (since `firstTry` is an argument to the function, not a member variable).

Having a `m_icons_loading` list (or set) would let us keep track of which ones we already have a download running for, but we'd need to somehow link that to the currently executing async function. This seems like obvious case where we should use Promises but I don't think they exist in Vala, and I'm not sure what the best alternative would be. Maybe exclusively using the signals and not have the `getIcon` function?

jangernert: Wouldn't it just load the existing metadata in that case and see that a download was already attemted recently?

brendanlong: Yes that! We can have `private Gee.HashMap<string, Gee.Future<Gdk.Pixbuf>> m_map;` then this should work:

```
public async Gdk.Pixbuf? getIcon(Feed feed)
{
    var feed_id = feed.getFeedID();
    var future = m_map.get(feed_id);

    if(future == null)
    {
        var promise = new Gee.Promise<Gdk.Pixbuf?>();
        future = promise.future;
        m_map.set(feed_id, future);
        load.begin(feed, (obj, res) => {
            try
            {
                promise.set_value(load.end(res));
            }
            catch(Error e)
            {
                promise.set_error(e);
            }
        });
    }

    // Assuming Gee.Future works properly, we should be able
    // to call these multiple times, and after the first time, it
    // should return instantly
    yield future.wait_async();
    return future.get_value();
}
```

jangernert: Would you like to commit these changes? Otherwise I'll try to take a look at it tonight.

brendanlong: Sure I'll try

brendanlong: Actually it's too complicated for me to do right now. I'll try to take a look tonight if you haven't already.

brendanlong: Almost have this working..

brendanlong:
@jangernert What do you think of these commits I added? In my testing it seems to work well, and first-time favicon loading is *way* faster. Icons seems to all show up in the right places and they show up when dragging.

I think this is good to land but I'm not sure if there's anything else you want checked @jangernert ?

jangernert:
The future is async I guess. This is still a pretty new approach to me that I have to get used to. But you seem pretty familar with all these concepts.
Just a few things I don't quite understand. You mentioned if the icon was already loading every other attempt to get the same icon from the manager would return instantly. But `return yield future.wait_async();` makes me think it will only return once the first loading process is finished and the future got assigned a value.
And reattempting to download an icon that failed to download before should work if I understand correctly. A nonexistant entry in `m_map` and an entry that failed to download before should return `null` so it will be downloaded again.
But lets say a FeedRow requested an icon and the manager failed to download it at that time. The UI is kept running and syncs are deactivated (I know this is unlikely). Now a few days later you click on that feed and load all its articles in the ArticleList. The ArticleRows will now query the icon from the manager again and this time the download succeeds. Now the ArticleRows will have the new favicon applied but the FeedRow won't. Or did I miss something important here?
I know that this will be no big deal most of the time because the FeedList is recreated after a sync and ArticleRows are dynamically deleted and recreated constantly as you scroll and change views.

brendanlong:
> You mentioned if the icon was already loading every other attempt to get the same icon from the manager would return instantly. But return yield future.wait_async(); makes me think it will only return once the first loading process is finished and the future got assigned a value.

Yes, you're right. It would only start one download, and then all of the waiting tasks would trigger at the same time when it finishes. Attempts to get an icon after it's downloaded will be instant though, since the `Future` already contains the value.

> And reattempting to download an icon that failed to download before should work if I understand correctly. A nonexistant entry in m_map and an entry that failed to download before should return null so it will be downloaded again.

It will only attempt to download an icon once per load right now, since if a download fails, the map still has an entry with a future containing `null` (so the future itself isn't null).

I think a future improvement would be putting the expires time in the map and redownloading it when it expires. I'll create a ticket for that.

Issue #576, opened 2017-10-20
Redownload icons in FaviconManager when they expire


brendanlong:
Right now we only attempt to download an icon once per FeedReader process:

1. When a sync starts, we check if the map contains a future for the icon. If there is, we wait and get its value.
2. If the map does not contain a future for the icon, we attempt to load it and set the future to either the pixmap or `null`
3. The download checks if the future is null, not if the contents are null, so we will never redownload an icon until we close and restart FeedReader

In general, the impact of this should be minimal, since we intentionally won't attempt to download an icon again for at least a week, but this could be improved:

 - Make the future contain both a pixmap and an expires time
 - When getting an icon, reload the icon if its not in the map *or if the expires time has passed*

Doing that would make us load the new icon for any newly displayed rows. To update existing rows:

 - Add a signal for updating the icon
 - Fire that signal every time we load an icon (after putting the new value in the map)
 - Hook up signal handlers in the places where we use feed icons (ArticleRow, FeedRow?).

jangernert:
Ah, the content of the future being `null` and the future itself being `null` got me  :D 
The part with the signal was already in place, but you removed it again. I though that was intentional^^
Okay so basically have the expiration timestamp with the future and readd the signal + handlers.
Doesn't seem too hard.

brendanlong:
Yeah, I removed the signal because my current version didn't actually use it (since we would never fire the signal more than once). I think we should add it back if we make this change.

jangernert:
#577

PR #577, opened 2017-10-21
Try to download favicons again after certain amount of time

jangernert:
It's 2am right now. So I am not very confident this is ready to merge. But in theory this should adress #576 and we can finally focus on other things than favicons :D

### DIFF
--- a/plugins/backend/feedbin/feedbinAPI.vala
+++ b/plugins/backend/feedbin/feedbinAPI.vala
@@ -39,10 +39,18 @@ public class FeedbinAPI : Object {
 		if(user_agent != null)
 			m_session.user_agent = user_agent;
 
-		m_session.authenticate.connect((msg, auth, retrying) => {
-			if(!retrying)
-				auth.authenticate(this.username, this.password);
-		});
+		m_session.authenticate.connect(authenticate);
+	}
+
+	~FeedbinAPI()
+	{
+		m_session.authenticate.disconnect(authenticate);
+	}
+
+	private void authenticate(Soup.Message msg, Soup.Auth auth, bool retrying)
+	{
+		if(!retrying)
+			auth.authenticate(this.username, this.password);
 	}
 
 	private Soup.Message request(string method, string path, string? input = null) throws FeedbinError
--- a/src/FavIconManager.vala
+++ b/src/FavIconManager.vala
@@ -17,13 +17,13 @@ public class FeedReader.FavIconManager : GLib.Object {
 
 	private struct FavIconData
 	{
-		public Gdk.Pixbuf? Icon;
-		public ResourceMetadata? Metadata;
+		public Gdk.Pixbuf? icon;
+		public ResourceMetadata? metadata;
 
 		public FavIconData(Gdk.Pixbuf? icon, ResourceMetadata? meta)
 		{
-			Metadata = meta;
-			Icon = icon;
+			this.metadata = meta;
+			this.icon = icon;
 		}
 	}
 
@@ -51,7 +51,7 @@ public class FeedReader.FavIconManager : GLib.Object {
 			var feed_id = feed.getFeedID();
 			var future = m_map.get(feed_id);
 
-			bool download_icon = future == null || (future.ready && (future.value.Metadata == null || future.value.Metadata.IsExpired()) );
+			bool download_icon = future == null || (future.ready && (future.value.metadata == null || future.value.metadata.is_expired()) );
 
 			if(download_icon)
 			{
@@ -87,7 +87,7 @@ public class FeedReader.FavIconManager : GLib.Object {
 			}
 
 			FavIconData data = yield future.wait_async();
-			return data.Icon;
+			return data.icon;
 		}
 		catch(Error e)
 		{
--- a/src/FavIconManager.vala
+++ b/src/FavIconManager.vala
@@ -15,6 +15,18 @@
 
 public class FeedReader.FavIconManager : GLib.Object {
 
+	private struct FavIconData
+	{
+		public Gdk.Pixbuf? Icon;
+		public ResourceMetadata? Metadata;
+
+		public FavIconData(Gdk.Pixbuf? icon, ResourceMetadata? meta)
+		{
+			Metadata = meta;
+			Icon = icon;
+		}
+	}
+
 	private Gee.Map<string, Gee.Future<FavIconData?>> m_map = new Gee.HashMap<string, Gee.Future<FavIconData?>>();
 	private static FavIconManager? m_cache = null;
 
--- a/src/FavIconManager.vala
+++ b/src/FavIconManager.vala
@@ -15,8 +15,10 @@
 
 public class FeedReader.FavIconManager : GLib.Object {
 
-	private Gee.Map<string, Gee.Future<Gdk.Pixbuf?>> m_map = new Gee.HashMap<string, Gee.Future<Gdk.Pixbuf?>>();
+	private Gee.Map<string, Gee.Future<FavIconData?>> m_map = new Gee.HashMap<string, Gee.Future<FavIconData?>>();
 	private static FavIconManager? m_cache = null;
+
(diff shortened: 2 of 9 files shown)


src/FavIconManager.vala
```diff
@@ -36,15 +38,19 @@ public class FeedReader.FavIconManager : GLib.Object {
 		{
 			var feed_id = feed.getFeedID();
 			var future = m_map.get(feed_id);
-			if(future == null)
+
+			bool download_icon = future == null || (future.ready && (future.value.Metadata == null || future.value.Metadata.IsExpired()) );
```
brendanlong: I'm not sure why we have a `null` check for `future.value.Metadata`. The `expires` could be null, but the metadata itself shouldn't ever be.

Also `future.value.Metadata.IsExpired()` seems to be an entirely new style of variable naming.. :p

src/FavIconManager.vala
```diff
@@ -36,15 +38,19 @@ public class FeedReader.FavIconManager : GLib.Object {
...
 				try
 				{
 					future = promise.future;
 					m_map.set(feed_id, future);
 
-					var stream = yield downloadFavIcon(feed);
+					ResourceMetadata metadata;
+					var stream = yield downloadFavIcon(feed, out metadata);
 					if(stream == null)
 						return null;
```
brendanlong: In this branch, we're going to set the `metadata` to null instead of using the returned `metadata`

src/Structs.vala
```diff
@@ -34,7 +34,19 @@ namespace FeedReader {
...
+	{
+		public Gdk.Pixbuf? Icon;
+		public ResourceMetadata? Metadata;
+
+		public FavIconData(Gdk.Pixbuf? icon, ResourceMetadata? meta)
+		{
+			Metadata = meta;
+			Icon = icon;
+		}
+	}
```
brendanlong: This should probably be defined in `FavIconManager` since it doesn't have any use outside of that class.

src/Structs.vala
```diff
@@ -131,6 +143,17 @@ namespace FeedReader {
...
+		public bool IsExpired()
+		{
+			if(expires == null)
+				return true;
+
+			if(expires.compare(new DateTime.now_utc()) == 1)
+				return false;
+
+			return true;
+		}
```
brendanlong: Nice :+1:

src/Widgets/ArticleRow.vala
```diff
@@ -267,18 +271,21 @@ public class FeedReader.ArticleRow : Gtk.ListBoxRow {
...
-		});
-
-		return icon;
+		var icon = yield FavIconManager.get_default().getIcon(feed);
+		if(icon != null)
+		{
+			if(inIcon == null)
+				m_icon.pixbuf = icon;
+			else
+				inIcon.pixbuf = icon;
```
brendanlong: I'd rather make `inIcon` a required argument and not have this branch.

src/Widgets/FeedRow.vala
```diff
@@ -152,19 +156,28 @@ public class FeedReader.FeedRow : Gtk.ListBoxRow {
...
+		if(inIcon == null)
+		{
+			m_icon.pixbuf = icon;
+			m_icon.get_style_context().remove_class("fr-sidebar-symbolic");
+		}
+		else
+		{
+			inIcon.pixbuf = icon;
+			inIcon.get_style_context().remove_class("fr-sidebar-symbolic");
+		}
```
brendanlong: Same thing as above: This condition seems way more complicated than just always passing in the icon.

src/Widgets/FeedRow.vala
```diff
@@ -44,9 +44,13 @@ public class FeedReader.FeedRow : Gtk.ListBoxRow {
 			var rowhight = 30;
 			m_box = new Gtk.Box(Gtk.Orientation.HORIZONTAL, 0);
 			m_icon = createFavIcon();
-
 			m_icon.margin_start = level * 24;
 
+			FavIconManager.get_default().ReloadFavIcon.connect(feed => {
+				if(m_feed.getFeedID() == feed.getFeedID())
+					reloadFavIcon.begin();
+			});
```
brendanlong: Do we need to manually disconnect this signal in the destructor? I'm not sure how Vala handles this

src/FavIconManager.vala
```diff
@@ -36,15 +38,19 @@ public class FeedReader.FavIconManager : GLib.Object {
...
-				var promise = new Gee.Promise<Gdk.Pixbuf?>();
+				var promise = new Gee.Promise<FavIconData?>();
 				try
 				{
 					future = promise.future;
 					m_map.set(feed_id, future);
 
-					var stream = yield downloadFavIcon(feed);
+					ResourceMetadata metadata;
+					var stream = yield downloadFavIcon(feed, out metadata);
```
brendanlong: This is only going to get the current metadata (what was  on the disk when the sync started. The new metadata isn't returned from `downloadFavIcon()` :(

brendanlong:
The version I just pushed made this a lot simpler by having a FavIcon class instead of a FavIconManager class. I also changed a few things so we only read and write metadata once, and writing the metadata in a finally clause gives us a the added benefit that if you open feedreader, then close it while it's downloading, we don't write the new metadata so we'll continue downloads that didn't finish (and this is per-icon, so we won't try to redownload icons that *did* finish).

brendanlong:
And I'm particularly happy about these functions:

```
	private Gtk.Image createFavIcon()
	{
		var icon = new Gtk.Image.from_icon_name("feed-rss-symbolic", Gtk.IconSize.LARGE_TOOLBAR);

		Feed feed = DataBase.readOnly().read_feed(m_article.getFeedID());
		var favicon = FavIcon.for_feed(feed);
		favicon.get_pixbuf.begin((obj, res) => {
			var pixbuf = favicon.get_pixbuf.end(res);
			if(pixbuf != null)
				icon.pixbuf = pixbuf;
		});
		favicon.pixbuf_changed.connect((feed, pixbuf) => {
			icon.pixbuf = pixbuf;
		});

		return icon;
	}
```

brendanlong:
I tried making the favicon cache have weak references, but it ended up throwing all of the favicons away every time it changed the interface (which is logical, since we presumably delete all of the old rows (at which point the favicons would be destroyed because nothing is referencing them), and then create a bunch of new icons). Ideally we'd have a way to eventually drop unused favicons but I doubt it's a huge problem so I'm leaving that alone for now.

plugins/backend/feedbin/feedbinAPI.vala
```diff
@@ -39,10 +39,18 @@ public class FeedbinAPI : Object {
...
 			m_session.user_agent = user_agent;
 
-		m_session.authenticate.connect((msg, auth, retrying) => {
-			if(!retrying)
-				auth.authenticate(this.username, this.password);
-		});
+		m_session.authenticate.connect(authenticate);
+	}
+
+	~FeedbinAPI()
```
bil-elmoussaoui: What does this line do? Should be removed i guess

brendanlong: I'm not sure which line you mean, but the destructor is disconnecting the `authenticate` callback to prevent issues where the callback is called after the object is destroyed (I'm not sure if it's really necessary in this case, but it's safer to do it than not).

jangernert:
What exactly do you mean by "every time it changed the interface" ?
I agree that deleting not used favicons is not all that important.

brendanlong:
> What exactly do you mean by "every time it changed the interface" ?

I mean like.. when we do a sync and then change the article rows. Any time we remove the old ones and add new ones.

jangernert:
Dropping some icons after a sync which don't have an article in the article list at the moment seems logical to me. But as soon as the feedlist is rebuilt every icon should have at least 1 reference in the feedlist (as long as you don't have show only unread activated). So changing the articlelist won't drop the ref count of any icon to 0.
Anyway, doesn't really matter right now :D

PR #579, opened 2017-10-22
Add Feedbin to the CI Docker image

brendanlong:
I never got a response from the owner of Feedbin, so this builds the whole project in our Docker image.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -4,8 +4,33 @@ jobs:
     docker:
       # See docker/Dockerfile
       - image: feedreader/fedora-feedreader-devel
+        environment:
+          ELASTICSEARCH_URL: http://127.0.0.1:9200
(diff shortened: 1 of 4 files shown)


jangernert:
So we build and run our own instance of feedbin every time? Feels like overdoing it. But if it's not slowing down things considerably: why not^^

brendanlong:
Luckily most of the work happens in the Docker build, so starting Feedbin is actually pretty fast (we start it in the background and it's ready by the time the build finishes). The main reason I did this is that our Feedbin trial ran out and I haven't heard back from the owner of Feedbin, so the options are (1) turn off the tests or (2) run our own Feedbin instance.

jangernert:
Funny how feedbin went from by far the worst back-end to one of the most tested in such a short time :)

Issue #578, opened 2017-10-21
error on add new feed


gutierri:
When I try to insert a new entry, the feedreader closes.

Debug:
> ** (feedreader:22976): DEBUG: UpdateButton: update status
** (feedreader:22976): DEBUG: Ping: status 200
** (feedreader:22976): DEBUG: Ping successful
** (feedreader:22976): DEBUG: Table CachedActions is empty
** (feedreader:22976): DEBUG: CachedActionManager - executeActions: no actions to perform
** (feedreader:22976): DEBUG: Ping: https://duckduckgo.com/
** (feedreader:22976): DEBUG: FeedReader: setOnline
** (feedreader:22976): DEBUG: Ping: status 200
** (feedreader:22976): DEBUG: Ping successful
** (feedreader:22976): DEBUG: FeedReader: updateSyncProgress
NULL
An error occured (SIGSEGV)
   The reason is likely a null reference being used.
   Note: no file path and line numbers can be retrieved. Are you sure you added -g -X -rdynamic to valac command line?
    # 1  <unknown>   in 'feed_reader_feed_server_syncContent'
        at feedreader
    # 2  <unknown>   in 'clone'
        at /usr/lib/libc.so.6

brendanlong:
Which version are you using, and which backend plugin (local, Feedly, TT-RSS, etc?)

gutierri:
From source and backend local

bil-elmoussaoui:
I had the same issue with local source too. Using the master branch, i didn't have the time to fix it locally yet

jangernert:
Does it happen with every feed you add? Any detail we should know about? I just tried it with a random rss feed and it worked fine for me.

bil-elmoussaoui:
i'm getting this issue whenever i try to sync the local RSS 
```shell
An error occured (SIGSEGV)
   The reason is likely a null reference being used.
   Note: no file path and line numbers can be retrieved. Are you sure you added -g -X -rdynamic to valac command line?

    #1  <unknown>   in 'atoi'
        at 
    #2  <unknown>   in 'feed_reader_feed_server_syncContent'
        at feedreader
    #3  <unknown>   in 'clone'
        at /usr/lib/libc.so.6
```

jangernert:
@brendanlong fixed this with #580

PR #580, opened 2017-10-23
Check if row ID is null before trying to parse as an int

brendanlong:
See issue #578


brendanlong:
(Need someone to test this before landing -- I can't build FeedReader on my current machine)

bil-elmoussaoui:
I will check that in few minutes 👍

jangernert:
Could reproduce the bug with an empty db. This patch fixes it.
Nice work!

PR #581, opened 2017-10-23
Only load all plugins before showin login page

jangernert:


src/Backend/FeedServer.vala
```diff
@@ -21,8 +21,11 @@ public class FeedReader.FeedServer : GLib.Object {
 	private FeedServerInterface? m_plugin;
 	private Peas.Engine m_engine;
 
+	public signal void PluginsChanedEvent();
```
brendanlong: Should be `PluginsChangedEvent`?

src/Backend/FeedServer.vala
```diff
@@ -56,10 +61,32 @@ public class FeedReader.FeedServer : GLib.Object {
...
+	public void LoadAllPlugins()
+	{
+		Logger.debug("FeedServer: load all available plugins");
 		foreach(var plugin in m_engine.get_plugin_list())
 		{
 			m_engine.try_load_plugin(plugin);
 		}
+
+		// have to readd this path, otherwise new icons won't show up
+		Gtk.IconTheme.get_default().add_resource_path("/org/gnome/FeedReader/icons");
```
brendanlong: That's odd. Why do we need the icon path to load plugins?

src/Widgets/LoginPage.vala
```diff
@@ -84,6 +72,35 @@ public class FeedReader.LoginPage : Gtk.Stack {
...
+				BackendInfo pluginfo = BackendInfo()
+				{
+					ID = plug.getID(),
+					name = plug.serviceName(),
+					flags = plug.getFlags(),
+					website = plug.getWebsite(),
+					iconName = plug.iconName()
+				};
+				m_accountList.add(new LoginRow(pluginfo));
+			}
```
brendanlong: Should we log if the cast to `FeedServerInterface` fails?

src/Backend/FeedServer.vala
jangernert: The plugins have logo icons compiled into their binaries as gresource. For some reason these icons can't be found if I don't add this path here again and only broken icons would show up on the login page.

src/Widgets/LoginPage.vala
jangernert: At least it wouldn't hurt.

PR #583, opened 2017-10-24
HIDPI favicons

jangernert:
Can be tested even without a hidpi display:
Last time I checked setting the scale factor in the gtk inspector doesn't work since it doesn't trigger to reload the icons. You need to start the app with `GDK_SCALE=2 feedreader`.

### DIFF
--- a/src/FavIcon.vala
+++ b/src/FavIcon.vala
@@ -37,14 +37,24 @@ public class FeedReader.FavIcon : GLib.Object
 	private Gee.Promise<Gdk.Pixbuf?> m_pixbuf = null;
 	private ResourceMetadata m_metadata;
 
-	public signal void pixbuf_changed(Feed feed, Gdk.Pixbuf pixbuf);
+	public signal void surface_changed(Feed feed, Cairo.Surface surface);
 
 	private FavIcon(Feed feed)
 	{
 		m_feed = feed;
 	}
 
-	public async Gdk.Pixbuf? get_pixbuf()
+	private int GetScaleFactor()
+	{
+		return MainWindow.get_default().get_style_context().get_scale();
+	}
+
+	private Cairo.Surface CreateSurfaceFromPixbuf(Gdk.Pixbuf pixbuf)
+	{
+		return Gdk.cairo_surface_create_from_pixbuf(pixbuf, GetScaleFactor(), null);
+	}
+
+	public async Cairo.Surface? get_surface()
 	{
 		// wait for ready + expired so we don't make a bunch of requests at once
 		if(m_pixbuf == null || (m_pixbuf.future.ready && m_metadata.is_expired()))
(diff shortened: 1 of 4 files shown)


src/FavIcon.vala
```diff
@@ -37,14 +37,24 @@ public class FeedReader.FavIcon : GLib.Object
...
 		m_feed = feed;
 	}
 
-	public async Gdk.Pixbuf? get_pixbuf()
+	private int GetScaleFactor()
+	{
+		return MainWindow.get_default().get_style_context().get_scale();
+	}
+
+	private Cairo.Surface CreateSurfaceFromPixbuf(Gdk.Pixbuf pixbuf)
```
brendanlong: Note: I thought we wanted names like `create_surface_from_pixbuf`?

src/FavIcon.vala
```diff
@@ -37,14 +37,24 @@ public class FeedReader.FavIcon : GLib.Object
...
+	{
+		return Gdk.cairo_surface_create_from_pixbuf(pixbuf, GetScaleFactor(), null);
+	}
```
brendanlong: Since these two functions are both only used in `get_surface()` it seems simpler to just put this in `get_surface()`: 
```
var scale = MainWindow.get_default().get_style_context().get_scale();
return Gdk.cairo_surface_create_from_pixbuf(pixbuf, scale, null);
```

jangernert: We also need the scale in L96, hence the function.

src/FavIcon.vala
jangernert: Oh okay, note taken.

src/FavIcon.vala
brendanlong: But the two uses are right next to each other in the same function. A variable seems to make more sense than a function?

jangernert: makes sense

brendanlong: Oh wait, we do call `CreateSurfaceFromPixbuf` twice.. so nevermind, your version makes sense :D

src/FavIcon.vala
```diff
@@ -37,14 +37,24 @@ public class FeedReader.FavIcon : GLib.Object
...
 
-	public async Gdk.Pixbuf? get_pixbuf()
+	private int get_scale_factor()
+	{
+		return MainWindow.get_default().get_style_context().get_scale();
+	}
+
+	private Cairo.Surface create_surface_from_pixbuf(Gdk.Pixbuf pixbuf)
+	{
+		return Gdk.cairo_surface_create_from_pixbuf(pixbuf, GetScaleFactor(), null);
```
brendanlong: I think you missed `GetScaleFactor` here

jangernert: Indeed I did.

PR #585, opened 2017-10-28
Make a better SQLite interface

brendanlong:
This wraps the low-level SQlite interface with one that handles parameter binding and error checking automatically.

### DIFF
--- a/src/DataBaseReadOnly.vala
+++ b/src/DataBaseReadOnly.vala
@@ -31,19 +31,6 @@ public class FeedReader.DataBaseReadOnly : GLib.Object {
 		Logger.error(@"dbErrorLog: $code: $msg");
 	}
 
-	// Generate enough placeholders for n parameters
-	private static string n_params(int n)
-	{
-		StringBuilder s = new StringBuilder.sized(n * 2 - 1);
-		for(int i = 0; i < n; ++i)
-		{
-			if(i != 0)
-				s.append(",");
-			s.append("?");
-		}
-		return s.str;
-	}
-
 	public void init()
 	{
 		Logger.debug("init database");
@@ -152,7 +139,7 @@ public class FeedReader.DataBaseReadOnly : GLib.Object {
 	public uint get_unread_total()
 	{
 		var query = "SELECT COUNT(*) FROM articles WHERE unread = ?";
-		var rows = m_db.execute(query, ArticleStatus.UNREAD.to_string());
+		var rows = m_db.execute(query, { ArticleStatus.UNREAD.to_string() });
 		assert(rows.size == 1 && rows[0].size == 1);
 		return (int)rows[0][0];
 	}
@@ -160,7 +147,7 @@ public class FeedReader.DataBaseReadOnly : GLib.Object {
 	public uint get_marked_total()
 	{
 		var query = "SELECT COUNT(*) FROM articles WHERE marked = ?";
-		var rows = m_db.execute(query, ArticleStatus.MARKED.to_string());
+		var rows = m_db.execute(query, { ArticleStatus.MARKED.to_string() });
 		assert(rows.size == 1 && rows[0].size == 1);
 		return (int)rows[0][0];
 	}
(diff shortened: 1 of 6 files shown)


src/DataBaseReadOnly.vala
```diff
@@ -15,289 +15,154 @@
...
-		while(stmt.step() == Sqlite.ROW)
-		{
-			for(int i = 0; i < cols; i++)
-			{
-				count = stmt.column_int(i);
-			}
-		}
-		stmt.reset();
-
-		return count;
```
brendanlong: Note: Several of these functions were unused

src/DataBaseReadOnly.vala
```diff
@@ -621,71 +347,58 @@ public class FeedReader.DataBaseReadOnly : GLib.Object {
...
-
-		Sqlite.Statement stmt;
-		int ec = sqlite_db.prepare_v2(query.get(), query.get().length, out stmt);
-		if(ec != Sqlite.OK)
-		{
-			Logger.error(query.get());
-			Logger.error(sqlite_db.errmsg());
-		}
+		var query = "SELECT articleID, unread, marked FROM articles WHERE articleId IN (%s)".printf(n_params(ids.size));
+		var rows = m_db.executev(query, ids);
```
bil-elmoussaoui: This should be execute instead of executev

brendanlong: In this one it's `executev` because `ids` is already a `List<string>`

bil-elmoussaoui: Ah alright, i just took a look using my phone and thought that it was a typo! should have been more careful :p

brendanlong: At this point the commit is *very* broken, so I'm mainly looking for high-level comments. It's not even close to building yet :)

brendanlong:
Hm to make this reasonable I'll probably need to normalize the categories, having a category table instead of a comma-separate string..

brendanlong:
Ok, I reduced the scope of the first commit by a lot, so it actually compiles now, and removes most of the duplicate code around creating prepared statements.

I'm not sure what the right thing to do with the `execute` function is though, since:

 - varargs aren't type safe. This commit assumes all arguments are strings, but it's very easy to accidentally pass an int
 - `Value` doesn't auto-convert when casting, so doing the obvious thing (always returning Variant with an `int64`) doesn't work, since trying to cast that to `int` will inexplicably fail.

brendanlong:
Looks like using an array argument is pretty ergonomic, since you just surround the parameters with `{}`. It would be even nicer if we could pass the `int` arguments as ints though..

brendanlong:
Ooo nice. Vala auto-casts to `Value`, so it's actually even more ergonomic to use a `Value[]` for the params.

Inside `execute`, it's not a big deal to do some checks and transformations, but I'm still not sure how to make the return values sane, since casting doesn't transform by default.

brendanlong:
Huh, so once I made Sqlite.Value work, auto-casting works perfectly there.

brendanlong:
Ahhh annoyingly it will let us cast, but `(int)some_sqlite_value` actually casts the *pointer* to an int. Ugh.

brendanlong:
Ok, so this works pretty well, and fixes most of the manual parameter handling. The main things left are:

 - We could make the inserts and updates a lot simpler, but it wasn't a priority since they're already using parameters
 - The last few read-only functions are really complicated and I'll probably want to tackle them one at a time
 - There are some places where we do multiple queries where it would be relatively easy to turn them into joins now, but I didn't want to deal with it in this commit

I think this is ready to merge as soon as it gets reviewed.

vapi/sqlite3.vapi
```diff
@@ -0,0 +1,460 @@
...
+		public int64 to_int64 ();
+		[CCode (cname = "sqlite3_value_text")]
+		public unowned string to_string ();
+		[CCode (cname = "sqlite3_value_type")]
+		public int to_type ();
+		[CCode (cname = "sqlite3_value_numeric_type")]
+		public int to_numeric_type ();
+		[CCode (cname = "sqlite3_value_dup")]
+		public Value copy ();
+	}
```
brendanlong: This file is an exact copy of the sqlite3.vala distributed by valac, except I added `copy()` / `sqlite3_value_dup()`, and set the `copy_function`, `free_function`, and `[Immutable]`, to make it work with `Gee.List`

brendanlong: Oh and I renamed `to_text()` to `to_string()`

jangernert:
Just tested it and it works great.
There are a few annoying compiler warnings about not handling the SQLiteError exception in SQLite.vala. Do you have any idea why that is? Handling the exception right there doesn't make much sense, but catching it in the calling function doesn't get rid of the warning for me.

brendanlong:
> There are a few annoying compiler warnings about not handling the SQLiteError exception in SQLite.vala. Do you have any idea why that is? Handling the exception right there doesn't make much sense, but catching it in the calling function doesn't get rid of the warning for me.

Oh right, I need to look at this more. The problem is that the Gee library defines the functions it takes wrong. The exceptions actually *are* handled, but the delegate declaration for `FutureMapFunc` is missing `throws Error`. We definitely don't want to catch the errors, since that would prevent them from bubbling up :(

PR #587, opened 2017-10-29
Better sync limit handling

### LINKED ISSUE #556: Don't download articles we're going to delete in inital sync - handle in a general purpose way
This is documenting something I've been working on off and on.

I've been using Feedbin since it first came out, and have 241 subscriptions, so an intial sync took *forever*. This was especially annoying because I had the "only keep one month" setting set, so it downloads for hours, and then deletes 99.9% of the data. I fixed this hackishly for Feedbin:

```
		var settings_state = new GLib.Settings("org.gnome.feedreader.saved-state");
		DateTime? time = null;
		switch(Settings.general().get_enum("drop-articles-after"))
		{
			case DropArticles.ONE_WEEK:
				time = new DateTime.now_utc().add_weeks(-1);
				break;

			case DropArticles.ONE_MONTH:
				time = new DateTime.now_utc().add_months(-1);
				break;

			case DropArticles.SIX_MONTHS:
				time = new DateTime.now_utc().add_months(-6);
				break;
		}
		if(!dbDaemon.get_default().isTableEmpty("articles"))
		{
			var last_sync = new DateTime.from_unix_utc(settings_state.get_int("last-sync"));
			if(time == null || last_sync.to_unix() > time.to_unix())
			{
				time = last_sync;
			}
		}
```

.. but ideally we'd do this in a way that all backends can use, without looking at the settings. I'm also looking back farther if we change the setting (for example, if you have several months of articles in Feedbin, change the setting to "only keep one week", then change the setting back to "forever", we'll never download your old articles again -- just keep new ones longer).

I might be able to handle this by passing in a `DateTime? since` argument and calculate that in FeedServer.

brendanlong:
This makes us pass a `since` argument to `getArticles()`, which:

 - Is a lot more efficient than what the Feedbin plugin was doing, since we only do it once per sync
 - Makes it easier for other plugins to re-use this logic (if any of them have an API like this)
 - Lets us easily do a full sync on newly added feeds (this was broken in the Feedbin plugin)

I also moved some things to the `DropArticles` enum to make them easier to reuse.

Fixes #556

### DIFF
--- a/src/Backend/FeedServer.vala
+++ b/src/Backend/FeedServer.vala
@@ -230,23 +230,9 @@ public class FeedReader.FeedServer : GLib.Object {
 			Notification.send(newArticles);
 		}
 
-		switch(Settings.general().get_enum("drop-articles-after"))
-		{
-			case DropArticles.NEVER:
-				break;
-
-			case DropArticles.ONE_WEEK:
-				DataBase.writeAccess().dropOldArtilces(1);
-				break;
-
-			case DropArticles.ONE_MONTH:
-				DataBase.writeAccess().dropOldArtilces(4);
-				break;
-
-			case DropArticles.SIX_MONTHS:
-				DataBase.writeAccess().dropOldArtilces(24);
-				break;
-		}
+		var drop_weeks = ((DropArticles)Settings.general().get_enum("drop-articles-after")).to_weeks();
+		if(drop_weeks != null)
+			DataBase.writeAccess().dropOldArtilces(-(int)drop_weeks);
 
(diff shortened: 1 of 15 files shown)


src/Backend/FeedServer.vala
```diff
@@ -962,12 +966,12 @@ public class FeedReader.FeedServer : GLib.Object {
 		return m_plugin.getUnreadCount();
```
brendanlong: I'm not really sure how to convert this one so I left it alone for now. Feedbin currently ignores this function anyway....

Issue #586, opened 2017-10-29
Replace ownCloud with Nextcloud


schiessle:
The maintainer of the news app decided to move to Nextcloud, see https://github.com/nextcloud/news. ownCloud no longer maintain/develop the app at all, seehttps://github.com/owncloudarchive

Therefore I would suggest to replace ownCloud with Nextcloud, both in the app and on the webpage.

jangernert:
@schiessle I would be happy to merge a pull request that does that.
Basically need new icons and change all the strings in the UI. Renaming the files and classes would be nice but is imo optional ;)

https://github.com/jangernert/FeedReader/tree/master/plugins/backend/owncloud

PR #589, opened 2017-10-30
Replace ownCloud with Nextcloud

schiessle:
ownCloud no longer maintains and support the News app. The News app is now part of Nextcloud.

@jangernert I only changed the strings and icons, as discussed in the issue. I didn't want to break anything and keep the upgrade path as smooth as possible.

fix https://github.com/jangernert/FeedReader/issues/586


brendanlong:
Looks good to me but @jangernert knows more about this stuff than I do.

It would be nice if the setup page somehow told people that the NextCloud plugin supports OwnCloud too. Maybe (as a future commit), we should make it so we can alias plugins, so we still present the OwnCloud plugin, but internally it's just the NextCloud plugin. I suspect we could use it for other plugins too, since a bunch of services support the Google API.

schiessle:
> It would be nice if the setup page somehow told people that the NextCloud plugin supports OwnCloud too.

I wouldn't mind if ownCloud would still support News. But if you look at the ownCloud app store you will see that they no longer offer the News app https://marketplace.owncloud.com/search/news. On github they moved the app to a organization which contain only dead/unmaintained code: https://github.com/owncloudarchive/news

So I don't know if it makes sense to still mention ownCloud if there isn't even a News app for recent ownCloud versions? But if you want to mention ownCloud somewhere, feel free to add it to a appropriated place.

brendanlong:
I was thinking that for people who already have it installed, we don't want to make it look like we dropped support. Just because you can't easily install it anymore doesn't mean everyone is going to instantly stop using it.

jangernert:
Looks good to me.

PR #595, opened 2017-11-05
GAction instead of deprecated Gtk.Action

jangernert:
Sadly I can't compile this yet. Fedora already ships webkitgtk 2.18. But the bindings were only updated in vala 0.38, which isn't shipped yet. What is the situation in other major distributions? Are webkit 2.18 and vala 0.38 in place?
We should imo wait a little bit to merge this until at least fedora catches up^^
Also have to update the custom vapi we ship.


src/Widgets/ArticleView.vala
```diff
@@ -773,11 +773,11 @@ public class FeedReader.ArticleView : Gtk.Overlay {
 		if(hitTest.context_is_image())
 		{
 			var uri = hitTest.get_image_uri().substring("file://".length);
-			var action = new Gtk.Action("save", _("Save image as"), null, null);
+			var action = new GLib.SimpleAction ("save", null);
 			action.activate.connect(() => {
 				Utils.saveImageDialog(uri);
 			});
-			menu.append(new WebKit.ContextMenuItem(action));
+			menu.append(new WebKit.ContextMenuItem.from_gaction(action, _("Save iamge as")));
```
brendanlong: Typo, should be "Save image as"

jangernert: thx

brendanlong:
Since we have a custom VAPI, I don't think we need to wait on Vala, we can just copy the VAPI from a newer version (I think?).

The current version of Ubuntu (including the LTS) supports WebKitGTK 2.18: https://packages.ubuntu.com/search?searchon=sourcenames&keywords=webkit2gtk

Debian stable is on 2.16 but has it in backports: https://packages.debian.org/source/stretch-backports/webkit2gtk

I would be shocked if Arch doesn't have it..

I don't have any strong opinion on whether we do this or now. I'm Fedora and don't really care about any other distro :p

jangernert:
Yes, copying the vapi would work. Just have to get it from somewhere^^

bil-elmoussaoui:
Vala and Webkit2Gtk are updated to the latest version on Archlinux stable repository!
```
╰─ pacman -Ss vala           
extra/vala 0.38.2-1 [installed]
    Compiler for the GObject type system
```
```
╰─ pacman -Ss webkit2gtk
extra/webkit2gtk 2.18.2-1 [installed]
    GTK+ Web content engine library
```

brendanlong:
@jangernert I copied the latest WebKitGTK VAPI and its dependencies, then got this building (had to set minimum Vala back to 0.36 so it would compile, and add to add an extra argument to that function we're calling now).

jangernert:
I just patched the webkit vapi to support that new method.

brendanlong:
Oh looks like you beat me to it :p

brendanlong:
Here's what I came up with: https://github.com/brendanlong/FeedReader/commit/c0fe6d2b92f0e113ef1f09fd87f9fa3dfda888f3

jangernert:
@brendanlong ah, was missing the jscore vapi.

Issue #593, opened 2017-11-04
Build error on Arch Linux


z3ntu:
```
[  4%] Built target vilistextum
[ 10%] Built target FeedReaderCommon
[ 11%] Built target gd
[ 11%] Built target gtkimageview
[...]

PR #622, opened 2018-01-12
Remove webkit2gtk vapi

jtojnar:
Introduced in aa6e6fba86f751004d1b5d52e6766a7b146abdae, it is not actually needed when the dependency is specified correctly – that is webkit2gtk-4.0, not WebKit2-4.0.

Closes: #593


jangernert:
Uh oh, looks like the CI server is not recent enough and is missing the webkit menu entry creation from gaction.
@brendanlong do we have to revert this after all, or can we somehow use a more recent distibution as build environment?

brendanlong:
I'll try moving us to Fedora 27

jtojnar:
Fedora 26 should contain webkitgtk 2.18.4, which compiles for me on NixOS.

jangernert:
@jtojnar I believe the vapi of webkitgtk is part of vala and so the vala version is the interesting here.

jtojnar:
It builds fine even with vala 0.36.8.

brendanlong:
Hmmm I can make a new image but the docker push isn't working:  #623

brendanlong:
Oh maybe I got it..

brendanlong:
Fixed with #623

jangernert:
Nice!

PR #634, opened 2018-01-28
Spelling: "article I used FeedReader to read"

comradekingu:
The idea here is that when it isnt "sent by FeedReader", the user has choice. "With" would do that, but this is way shorter.


jangernert:
Makes sense. Thx

PR #649, opened 2018-03-22
Fix app menu in GNOME

### LINKED ISSUE #647: Wrong placement of Application Menu
In GNOME Shell, when Application Menu on the top panel is disabled (using GNOME Tweak Tool), an "Application" menu appeared just below the header bar, which is redundant, as the same menu can be access with the app icon at the top left-hand corner of the header bar.

screenshot from 2018-03-21 10-56-56 screenshot from 2018-03-21 10-13-59

Comparing this to the screenshot [here](https://raw.githubusercontent.com/jangernert/feedreader/gh-pages/images/gallery/Screenshot4.png), it appears the menu button has disappeared from the right of the search box.
screenshot from 2018-03-21 11-37-47

peterychuang:
Sorry for sending and closing another pull request yesterday. I realised that fix wasn't correct.
Tested on GNOME and KDE on Arch Linux

Fix #647

### DIFF
--- a/src/Widgets/MainWindow.vala
+++ b/src/Widgets/MainWindow.vala
@@ -41,7 +41,7 @@ public class FeedReader.MainWindow : Gtk.ApplicationWindow
 
 	private MainWindow()
 	{
-		Object(application: FeedReaderApp.get_default(), title: _("FeedReader"));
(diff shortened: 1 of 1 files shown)


jangernert:
Nice, didn't know that property existed.

PR #650, opened 2018-04-03
Fixed syntax error in ttrss-plugin

mluettecke:
Just noticed the missing when I wanted to install the plugin.


jangernert:
thx

mluettecke:
Thanks to you for feedreader

jplatte:
It's not a syntax error, it's a best practice: https://stackoverflow.com/questions/4410704/why-would-one-omit-the-close-tag

PR #654, opened 2018-04-20
Fix parsing date on non English locales

### LINKED ISSUE #652: Date of feed items are not detected on some locales
Using the local backend, the date of the feed items are not detected, and always the current time is shown.

Steps to reproduce:

1. Start feedreader with German or Hungarian locale, e.g.: `LANG=de_DE.UTF-8 feedreader`
2. Switch to the local backend.
3. Add some feeds.
4. See that the date of each items are the current time instead of the date provided by the feed author.

Tested with the latest git version as of 2018.04.19.

City-busz:
Fixes #652

### DIFF
--- a/plugins/backend/local/localInterface.vala
+++ b/plugins/backend/local/localInterface.vala
@@ -603,7 +603,10 @@ public class FeedReader.localInterface : Peas.ExtensionBase, FeedServerInterface
 					if(item.pub_date != null)
 					{
 						GLib.Time time = GLib.Time();
+						var savedlocale = GLib.Intl.setlocale(GLib.LocaleCategory.ALL);
+						GLib.Intl.setlocale(GLib.LocaleCategory.ALL, "C");
 						time.strptime(item.pub_date, "%a, %d %b %Y %H:%M:%S %Z");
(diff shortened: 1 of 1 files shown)


City-busz:
> I was hoping to find an RFC 822 date parser but I can't find a pre-existing one. If I have time I'll write to write one in Vala.

That would be great. The currently used [GLib.Time](https://valadoc.org/glib-2.0/GLib.Time.html) ignores timezone, so the stored time is shifted.

brendanlong:
See #658

brendanlong:
@City-busz This should be fixed by 78d7ced862aa491d64d9b5f135b2ba2d893b79d1. Let me know if you have any other problems with this (or test cases to add: https://github.com/jangernert/FeedReader/blob/master/plugins/backend/local/TestLocalRSS.vala)

City-busz:
It works as expected. Thanks!

However, it breaks parallel build (make -j8) somehow:
`make[2]: *** No rule to make target 'FeedReader.vapi', needed by 'plugins/backend/local/local_target_valac.stamp'.  Stop.`

PR #659, opened 2018-04-21
Nulls and asserts

brendanlong:


brendanlong:
I don't need a review on this; just opening the pull request so the CI tests will run before merging.

PR #662, opened 2018-04-22
Meson

brendanlong:
Another attempt at switching to meson.

 - Rebased
 - Tested on my machine, switching libFeedReader to a dynamic lib fixed the crashes for me
 - Turn on tests

Unfortunately we're still getting weird glib errors with the Feedbin tests, but I can confirm that they don't occur when running `feedreader` normally, so I think we should just switch and then fix that.


brendanlong:
I asked about this on the GTK mailing list to see if we can get some help: https://mail.gnome.org/archives/gtk-list/2018-April/msg00009.html

brendanlong:
Another weird thing is that we link a main function into our shared lib. I wonder if that's also confusing something.

brendanlong:
Fixing the main function didn't help :(

jangernert:
4th iteration by the 3rd person and we're finally there.
If the workaround with a shared lib works it's fine imo. I'm wondering how cmake did it internally. Like I said there I just made sure the library and vapi was generated before the plugins were built and just specified the vapi file.
I just tested the branch in a freshly installed fedora VM and everything seems to work fine so far.
Only thing I noticed is, 'libunity' seems to be a hard dependency now. Did I do this? I thought I had it fixed by the end of my meson work.

brendanlong:
When I run it, meson says:

```
[...]
Dependency unity found: NO
Configuring Constants.intermediate.vala using configuration
Library webkit2gtk-web-extension-4.0 found: YES
Native dependency glib-2.0 found: YES 2.54.3
Library gd-1.0 found: YES
Library gtkimageview found: YES
Library libvilistextum found: YES
Library m found: YES
Message: libUnity support OFF
[...]
```

So I think this is still optional?

jangernert:
No idea what happened there yesterday. I thought I remembered an error about no libunity being found. But I checked just now again. And everything works as expected. Once `libunity` was removed `meson` completed successfully and reported `libunity` as OFF.
Maybe I had a different lib missing and just saw the red text next to `unity` and jumped to conclusions early.
If you want to go ahead and merge :)

bilelmoussaoui:
YEES! Finally we got the meson build system working :D it was really a hard task than I thought

bilelmoussaoui:
I think that we should really release a 2.1 version that includes all those small fixes we have already on master + the switch to meson build system! Once we have done that, we can work on the rest of the issues that should be there for the 2.2. 
I will try to finish the new themes too for the 2.2

PR #667, opened 2018-04-29
Enclosure and taggings

jangernert:
- getting rid of comma separated strings
- fix enclosure button always showing

and unrelated small fixes I did along the way
- copy GrabberConfig folder on install
- fix path to GrabberConfig

Not sure if I missed something. But that's all I could squeeze into one weekend.

### DIFF
--- a/plugins/backend/ttrss/ttrssAPI.vala
+++ b/plugins/backend/ttrss/ttrssAPI.vala
@@ -481,7 +481,7 @@ public class FeedReader.ttrssAPI : GLib.Object {
 					}
 				}
 
-				Gee.List<string>? media = null;
+				var enclosures = new Gee.ArrayList<Enclosure>();
 				if(headline_node.has_member("attachments"))
 				{
 					var attachments = headline_node.get_array_member("attachments");
@@ -490,16 +490,23 @@ public class FeedReader.ttrssAPI : GLib.Object {
 					if(attachments != null)
 						mediaCount = attachments.get_length();
 
-					if(mediaCount > 0)
-					 	media = new Gee.ArrayList<string>();
 					for(int j = 0; j < mediaCount; ++j)
 					{
 						var attachment = attachments.get_object_element(j);
-						if(attachment.get_string_member("content_type").contains("audio")
-						|| attachment.get_string_member("content_type").contains("video"))
+						EnclosureType type = EnclosureType.FILE;
+						if(attachment.get_string_member("content_type").contains("audio"))
 						{
-							media.add(attachment.get_string_member("content_url"));
+							type = EnclosureType.AUDIO;
 						}
+						else if(attachment.get_string_member("content_type").contains("video"))
+						{
+							type = EnclosureType.VIDEO;
+						}
+
+						enclosures.add(new Enclosure(
+							headline_node.get_string_member("id"),
+							attachment.get_string_member("content_url"),
+							type));
 					}
 				}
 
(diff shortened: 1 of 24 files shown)


plugins/backend/ttrss/ttrssAPI.vala
```diff
@@ -490,16 +490,23 @@ public class FeedReader.ttrssAPI : GLib.Object {
...
+						EnclosureType type = EnclosureType.FILE;
+						if(attachment.get_string_member("content_type").contains("audio"))
 						{
-							media.add(attachment.get_string_member("content_url"));
+							type = EnclosureType.AUDIO;
 						}
+						else if(attachment.get_string_member("content_type").contains("video"))
+						{
+							type = EnclosureType.VIDEO;
+						}
```
brendanlong: I wonder if we should add an `EnclosureType.from_string` function to do this?
```
static EnclosureType from_string(string str)
{
  if (str.contains("audio"))
    return AUDIO;
  else if (str.contains("video"))
    return VIDEO;
  else
    return FILE;
}
```
?

src/DataBaseReadOnly.vala
```diff
@@ -759,6 +794,34 @@ public class FeedReader.DataBaseReadOnly : GLib.Object {
...
 
+	private Gee.List<string> read_taggings(string articleID)
+	{
+		var list = new Gee.LinkedList<string>();
+
+		var rows = m_db.execute("SELECT * FROM taggings WHERE articleID = ?", { articleID });
+
+		foreach(var row in rows)
+		{
+			list.add(row[1].to_string());
```
brendanlong: Since we're only using one thing from the row, it would probably be better to explicitly list the column we want instead of doing a `SELECT *`

src/DataBaseReadOnly.vala
```diff
@@ -759,6 +794,34 @@ public class FeedReader.DataBaseReadOnly : GLib.Object {
...
+		}
+
+		return list;
+	}
+
+	private Gee.List<string> read_taggings_reverse(string tagID)
```
brendanlong: Consider renaming these functions `read_taggings_by_article_id` and `read_taggings_by_tag_id`

src/Model/Enclosure.vala
```diff
@@ -0,0 +1,38 @@
...
+
+	public string get_url()
+	{
+		return m_url;
+	}
+
+	public EnclosureType get_enclosure_type()
+	{
+		return m_type;
+	}
```
brendanlong: Consider using getter properties instead of functions.

src/StringUtils.vala
```diff
@@ -36,4 +36,14 @@ public class FeedReader.StringUtils {
 	{
 		return string.joinv(sep, l.to_array());
 	}
+
+	public static Gee.List<string> encapsulate(Gee.List<string> l, string encapsulation)
+	{
+		for(int i = 0; i < l.size; i++)
+		{
+			l[i] = encapsulation + l[i] + encapsulation;
```
brendanlong: If this is for quoting SQL strings, we need to also double all single quotes:

> A string constant is formed by enclosing the string in single quotes ('). A single quote within the string can be encoded by putting two single quotes in a row - as in Pascal. C-style escapes using the backslash character are not supported because they are not standard SQL.

(Also consider moving this to a static function on the Sqlite class)

```
public static string quote_string(string str)
{
    var escaped = str.replace("'", "''");
    return @"'$(escaped)'";
}
```

PR #670, opened 2018-05-05
Do a 2.1 (or 3.0) release

brendanlong:
I'll tag this once we merge.

Do I need to do anything in particular to make Flatpak work?

I can write up a release notes thing.

PR #672, opened 2018-05-06
update flatpak build

bilelmoussaoui:
This PR updates the flatpak dependencies to the latest versions in order to make it work with GNOME runtime 3.28, adds two tests for validating the desktop and the appdata file. Also updates the README & INSTALL build instructions.

This also removes the flatpak directory as we don't need to have our own repository anymore. We don't release daily builds (I guess?). The stable releases are now built using flathub so we don't have to carry that anymore.


jangernert:
@jscurtu 
Yes, the daemon d-bus name should not be necessary any more. And gnome-online-accounts 3.28 should be sufficient.

bilelmoussaoui:
- goa 3.29 is a stable version (I guess)
- the branch tag is not needed (I guess). It builds from master as I would like to add a pipeline that builds a flatpak package for each PR. People who wants stable releases won't even try to build the flatapk packag manually and they will just use flathub.
- The daemon was merged

brendanlong:
Does that mean we don't need to bother updating the stable branch as in #670 ?

bilelmoussaoui:
We don't need to update stable aymore as the stable releases are tagged. But we can tag the stable branch and provide hotfixes on that branch :+1:

bilelmoussaoui:
Downgraded goa to 3.28.

bilelmoussaoui:
The build fails due to some issues with ruby gems being not able to install needed dependencies for tests that were not modified in this PR ! so it's can be merged :+1:

PR #703, opened 2018-06-09
Solus OS is now just Solus and feedreader is now updated in the repos

Jacalz:
Feedreader has since been updated and is available without any issues in the repo :)


Jacalz:
You are welcome, thanks for merging 👍

PR #705, opened 2018-06-17
FreshAPI: Fix broken editTag function

### LINKED ISSUE #699: Article read status not synced back to FreshRSS
Problem; the articles I have read is changing status to unread when syncing with FreshRSS, possibly indicating a failure to sync changes back to the server. Have anyone else experienced this?
On my Android phone I use FeedMe, and the sync works as expected there.

klausenbusk:
The freshAPI editTag function was incorrectly using the wrong post field
to specify articleID ("r" instead of "i").
Which meant you couldn't (un)read/(un)star a article.

Fix #699

### DIFF
--- a/plugins/backend/fresh/freshAPI.vala
+++ b/plugins/backend/fresh/freshAPI.vala
@@ -292,7 +292,7 @@ public class FeedReader.freshAPI : Object {
 
 		foreach(string id in arrayID)
 		{
-			msg.add("r", "-/" + id);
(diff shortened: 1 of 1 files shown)


jangernert:
Just checked the [fresh API](https://github.com/FreshRSS/FreshRSS/blob/master/p/api/greader.php#L876). You're right. Don't know how nobody noticed it till now.

Issue #733, opened 2018-08-23
"sync" thread is hung at function feed_reader_utils_UTF8fix in Utils.vala:117 with string "And we\031re almost home&"


Raulvo:
During the GDB session I have seen that FeedReader creates threads while parsing the feeds. Thread 25 never exits and is in what to me seems an infinite loop within vilistextum trying to convert the string "And we\031re almost home&". FeedReader GUI remains in syncing state but there is no progress.

List of threads and backtrace of the hung thread:

```
(gdb) info threads
  Id   Target Id         Frame 
  1    Thread 0x7ffff7f8cac0 (LWP 10271) "feedreader" 0x00007ffff72e5bf9 in __GI___poll (fds=0x555555e9d7d0, nfds=4, timeout=9082) at ../sysdeps/unix/sysv/linux/poll.c:29
  2    Thread 0x7fffe0c2e700 (LWP 10275) "feedreader" 0x00007ffff0ec69f3 in futex_wait_cancelable (private=<optimised out>, expected=0, 
    futex_word=0x7fffea9dc1e0 <bmalloc::SafePerProcessStorageTraits<bmalloc::Scavenger>::Storage::s_memory+64>) at ../sysdeps/unix/sysv/linux/futex-internal.h:88
  3    Thread 0x7fffdfda3700 (LWP 10276) "gmain" 0x00007ffff72e5bf9 in __GI___poll (fds=0x5555557bd890, nfds=2, timeout=-1) at ../sysdeps/unix/sysv/linux/poll.c:29
  4    Thread 0x7fffdf5a2700 (LWP 10277) "gdbus" 0x00007ffff72e5bf9 in __GI___poll (fds=0x7fff7c001610, nfds=4, timeout=-1) at ../sysdeps/unix/sysv/linux/poll.c:29
  5    Thread 0x7fffdeb94700 (LWP 10278) "dconf worker" 0x00007ffff72e5bf9 in __GI___poll (fds=0x555555a15870, nfds=1, timeout=-1) at ../sysdeps/unix/sysv/linux/poll.c:29
  6    Thread 0x7fffdccff700 (LWP 10279) "StorageManager" 0x00007ffff72e5bf9 in __GI___poll (fds=0x7fff88002650, nfds=1, timeout=-1) at ../sysdeps/unix/sysv/linux/poll.c:29
[...]

Issue #715, opened 2018-07-13
100% CPU usage when quitting the application


rivetchip:
Reporting 100% CPU usage on application quit

**How to reproduce :** local feeds, navigate through feeds, mark all as read, then quit the application.
**OS :** Fedora 27 x64
**Version :** 2.4-dev ( fresh install from sources, Master branch )

If you need further data, please let me know :smile: 

Capture from Gnome Monitor :

![capture d ecran de 2018-07-13 22-56-47](https://user-images.githubusercontent.com/7000625/42713713-b636db44-86f0-11e8-9f03-2b59ef88d509.png)

rivetchip:
Yep, seems fixed in 2.6

PR #736, opened 2018-08-25
Libvilistextum fuzzing

brendanlong:
We've been having a lot of trouble with libvilistextum but weren't able to reproduce it, so I decided to run it through american fuzzy lop and see if it could help.

It immediately found three infinite loops:

 - If an input ends with `&` or `<`, we will loop by reading the character and then pushing it back (fixed by "Don't put back EOF")
 - If there's an unclosed HTML tag, we loop infinitely on the EOF (fied by "Fix infinite loop in HTML tags")

AFL also found a few crashes but I wasn't able to reproduce them in gdb.


brendanlong:
Related to #733, and possibly others (#715)

brendanlong:
There's a crash if you have an HTML attribute with a very specific sequence of characters still. I think it has something to do with it being an invalid UTF-8 sequence but I haven't figured it out yet.

brendanlong:
I put notes about how I ran this here: https://github.com/jangernert/FeedReader/blob/master/FUZZ_TESTING.md

jangernert:
wow, this came out of nowhere. :+1:

PR #740, opened 2018-08-26
Fix tests

bilelmoussaoui:
- appstreamcli fails to detect oars rating, keep only appstream-util

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -46,7 +46,7 @@ jobs:
       - run:
           name: Test
           command: cd builddir && ninja test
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
@brendanlong  I couldn't manage to make the tests related to afl works correctly. It seems Ninja stops the tests after 30seconds.

bilelmoussaoui:
20minutes is still not enough for afl :/

bilelmoussaoui:
Already fixed that!

Issue #712, opened 2018-07-10
Suddenly encoding has broken


asm0dey:
I've used FeedReader for months and suddenly encoding has broken. Not there are lots of unprintable characters and I can't open anything.

![feedreader_037](https://user-images.githubusercontent.com/1297282/42491963-ce11ea84-841f-11e8-800d-814c13a98b49.png)

asm0dey:
If you need further data — please let me know.

brendanlong:
This might be the same issue as #677? Can you try the latest commit on master and see if it helps?

asm0dey:
@brendanlong I've just tried, nothing changed.

asm0dey:
More info:
`feedreader-git` aur package is broken is broken
flatpak version also has broken symbols

City-busz:
I confirm this bug on Arch Linux with version 2.2 (https://github.com/jangernert/FeedReader/commit/61f6331e700421ca917a309df9efa2e119890ba5) and with master (https://github.com/jangernert/FeedReader/commit/0d4342d2547dba65609ecd23797ad5e0946f58b8).

City-busz:
It seems that it's always a random string, maybe caused by an uninitialized variable somewhere.

asm0dey:
Who can find working revision?

brendanlong:
Is this problem only happening on Arch? I don't use it but maybe there's a different compiler version / library version?

asm0dey:
@brendanlong faild on on flatpak also

TheoLeCalvar:
Using @Stunkymonkey PKGBUILD worked fine for me.
No more weird encoding issues.
I'm on commit [33c73c63](https://github.com/jangernert/FeedReader/commit/33c73c63e34b21478ed00995255404ac73c60b4a).

RorraVox:
me too, annoyng bug on arch linux

RorraVox:
I've tried also to compile PKGPUILD that @Stunkymonkey have modified but nothing, same strage encoding (i think) error.

RorraVox:
I get this output:

feedreader-INFO: 17:31:40.285: FeedReader 2.2-dev
(feedreader:29257): Gtk-DEBUG: 17:31:40.423: Connecting to session manager
feedreader-INFO: 17:31:40.510: Plugin Name: Local RSS Backend
feedreader-INFO: 17:31:40.510: Plugin Version: 0.1
[...]

PR #746, opened 2018-08-29
Get rid of vilistextum

brendanlong:
See #712

Even with all of the improvements to libvilistextum, it still doesn't work right for several locales. This is a WIP to just remove vilistextum and replace it with glib's string.make_valid and [Gumbo](https://github.com/google/gumbo-parser/blob/master/examples/clean_text.cc).


jangernert:
What baffles me is that these problems appeared for a lot of people only recently
I guess `make_valid()` doesn't handle html entities. Also vilistextum was used to convert html to plain text previews. Before I used html2text. But that was forgiving regarding invalid html and required the html to be written to disk. Then the html2text binary was spawned to parse the file.

edit: ah I see you're using "gumbo" (never heard of it)
does gumbo also work on *mostly* plain text? like I mentioned quite a lot of feeds put html entities in their headlines or summaries.

brendanlong:
Yeah, I have no idea why libvilistextum stopped working. It seems like the problems started when we switched to meson, but we didn't change libvilistextum at that time. We did do a new release though, so I wonder if a lot of people were on much older versions?

I was trying to fix libvilistextum itself, but it seemed simpler to just replace it with two well-known libraries for the same thing.

I'll make sure Gumbo works for the HTML tag stripping before I merge this.

jangernert:
I think the [planetKDE](https://planet.kde.org/rss20.xml) feed has html entities instead of ampersands in their titles. But most of the feeds I read behave nicely.

```
<title>KDAB at SIGGRAPH &#x2013; 2018</title>
```

btw: making steady progress with the [rust lib](https://gitlab.com/news-flash/news_flash_base). So if you get tired dealing with the poor decisions I made in the past in FeedReader (like vilistextum) you could hop over and prevent me from doing more of them in the new code base :D

brendanlong:
It seems to handle this feed correctly:

screenshot from 2018-08-29 13-27-06

I was having issues with it for a while until I realized the local RSS plugin is apparently producing invalid characters (it's converting en dash to \x0013 when it should be \x2013). I'm looking into that piece but the rest is good..

brendanlong:
I don't understand what's wrong with the local plugin. For some reason the RSS parser is turning `&#x2013;` into `\x00\0x13` instead of `\x20\x13`. I can confirm that it was doing this before I replaced vilistextum though:

screenshot from 2018-08-29 13-51-25

jangernert:
I wouldn't worry about it right now.

brendanlong:
I fixed the flatpak and prepared a new release. I'll merge when the build finishes.

asm0dey:
@brendanlong I've found exact commit where hyerogliphs started appearing. On same DB one version doesn't show them and next commit shows. I don't know if there was something about library changes.

brendanlong:
@asm0dey What commit was that?

asm0dey:
@brendanlong https://github.com/jangernert/FeedReader/issues/712#issuecomment-412995514

asm0dey:
@brendanlong https://github.com/jangernert/FeedReader/issues/712#issuecomment-414050816

brendanlong:
@asm0dey Are you still running into the encoding issue (or a related issue)?

asm0dey:
@brendanlong nop, as I've witten in #712 :) And thank you again

PR #751, opened 2018-08-30
Exchanging get_dbh() with get_pdo() fixes #750

### LINKED ISSUE #750: tt-rss plugin throws a notice about legacy database
If I install and run the `api_feedreader` plugin, tt-rss throws…

`Notice: Legacy connect requested to mysql in /home/www/ttrss/classes/db.php on line 22`

According to [this thread](https://discourse.tt-rss.org/t/pdo-is-coming-heres-what-you-need-to-know/689/63) this message from _tiny tiny rss_ is intended to nag plugin developers for not using PDO to connect to the database. Couldn't fix it myself. 😢

pixolin:
If I get it right, to fix #750 we only need to replace `get_dbh()` with the new function `get_pdo()`.
(see [file classes/pluginhost.php](https://git.tt-rss.org/fox/tt-rss/src/master/classes/pluginhost.php#L96))

### DIFF
--- a/data/tt-rss-feedreader-plugin/api_feedreader/init.php
+++ b/data/tt-rss-feedreader-plugin/api_feedreader/init.php
@@ -20,7 +20,7 @@ class Api_feedreader extends Plugin {
 	function init($host)
 	{
 		$this->host = $host;
-		$this->dbh = $host->get_dbh();
+		$this->dbh = $host->get_pdo();
 		$this->host->add_api_method("addLabel", $this);
 		$this->host->add_api_method("removeLabel", $this);
 		$this->host->add_api_method("renameLabel", $this);


brendanlong:
It looks like this might work but I can't really review it. Does it work if you use this on your server?

pixolin:
I tested it on a server with the current release of _tiny tiny rss_ and after switching the function it works like a charm. No notices regarding using a legacy connection and _FeedReader_ connects to the server immediately.

jangernert:
I'm neither very familiar with the tt-rss code nor with php in general. I wrote the plugin just browsing through the code taking bits and pieces and putting them back together.
So all I can say is the change sounds reasonable and if it works for you it is probably fine. As far as I understand tt-rss is on kind of a rolling release anyway nowadays. So compatibility shouldn't be the biggest issue here.

brendanlong:
It sounds like this might be causing problems in #752

PR #768, opened 2018-10-09
Set max-width of page elements to 100%

### LINKED ISSUE #767: Large images make the page scroll horizontally
From this feed: https://blogs.gnome.org/aday/feed/

``
screenshot from 2018-10-09 10-42-50

But if you look at the page itself:

https://blogs.gnome.org/aday/2018/10/09/farewell-application-menus/

screenshot from 2018-10-09 10-44-05

So we're not properly forcing the images to scale down, even though the source page is. I think we can do this through some combination of width and max-width (maybe just set `max-width: 100%` on everything?)

brendanlong:
This causes extra-wide content to scale down.

Fixes #767

### DIFF
--- a/data/ArticleView/style.css
+++ b/data/ArticleView/style.css
@@ -55,6 +55,10 @@ img {
     height: auto;
 }
 
+* {
+    max-width: 100%;
+}
+
 ol,
 ul {
     list-style: disc outside none;


brendanlong:
Before:

screenshot from 2018-10-09 10-42-50

After:

screenshot from 2018-10-09 11-02-20

jangernert:
Ah, apparently we did only set `max-width` for img tags inside of figures.
Also great PR in the sense that we can reuse all the HTML & CSS for the new code base =)

PR #776, opened 2018-10-23
Promote Flatpak package using the Flathub badge

bilelmoussaoui:
Remove the instructions as they are already mentioned on the Flathub website. Installing Flatpak now grabs the portals too.


README.md
```diff
@@ -11,25 +13,6 @@ FeedReader is a modern desktop application designed to complement existing web-b
...
-Besides installing the Flatpak Framework, you should also install the following portal packages using your distributions package manager:
-
-  - `xdg-desktop-portal`
-  - `xdg-desktop-portal-gtk`
-
-Follow the Flatpak [setup guide](https://flatpak.org/setup/) for your distribution before installing FeedReader. You must ensure Flatpak is installed and add the Flathub repo. Then you can run:
-
-```
-flatpak install flathub org.gnome.FeedReader
-```
```
brendanlong: It might be good to leave at least this much:

> The Flatpak build works on any distro and will always track the newest release.
> For that reason, it's the only way we recommend that you install FeedReader (we
> really don't have the resources to support multiple distro-specific package).
>
> See [Flathub for installation instructions](https://flathub.org/apps/details/org.gnome.FeedReader)

I'm just worried the icon is only helpful for people who already know that they want the Flatpak version, but some people are probably going to come here and not know what Flatpak is or know what it is but not realize it's the preferred way to install FeedReader.

Issue #694, opened 2018-05-21
app stays in memory no option to change behaviour


chriscamacho:
The application stays in memory without anyway to exit it other than killing the process
while some like notification spam, it would be nice to have the option to actually exit the application rather than just sleep...
after using the close widget (top right of window) I see

$ ps x | grep feedreader
 1822 ?        SLl    1:14 feedreader

chriscamacho:
thought this was a feature without options...
however using ctrl-q (instead of windows close widget) I do not see any sleeping processes...

chriscamacho:
forget Gnome (most have) NO application should background itself like this.

X == close !

A much better option would be to have (an optional) feature - minimise to system tray, this actually isn't that hard to implement (wish I had the time) but see...

as a work around I'll be running it in alltray....

brendanlong:
If you're interested in writing a patch:

 - I will approve a pull request to add this to the settings (on the Interface tab)
 - We'd have to check with @jangernert and @bilelmoussaoui if we want to change the default (I'm not sure what people expect here)

jangernert:
A dconf-switch would be fine imo and a switch in the settings also okay. But I'm not sure about turning it on by default. A system tray icon is definitely not the way to go. There is a reason most desktops want to move away from that concept.

bilelmoussaoui:
Currently, GNOME doesn't have any specified documentation on how to deal with applications that can't stay open even if the window is closed. Polari has implemented a nice behavior with a pop up when the user closes the application to ask them if they want the application to keep running in the background or not. But it's still not something widely used by other GNOME apps. 

I would suggest implementing the same thing as Polari does and provide an easy way to change that behavior from the settings

chriscamacho:
outside gnome, there is no way to bring the window back, even clicking on the notification doesn't bring the window back, running the app again however does seem to bring it back, at least there doesn't seem to be multiple processes...

Issue #522, opened 2017-08-03
Minimize to tray


ToeiRei:
I would love to have FeedReader running in the background minimized to the tray to have it out of the way until some news appear (notification popping up)

bil-elmoussaoui:
The FeedReader Daemon (the binary that grabs your latest articles) is already running on the background. There's no need to keep the window running too.
FeedReader also follows some of elementary & GNOME HIG. And for both, tray icons are bad design decisions. 
So, this will be a won't fix

qazip:
How is having a tray icon for an app that is minimized a bad design decisions? I much rather have it, so I know that something is running in the background, than not having and having annoying popups.

Some window managers (for example i3wm) do not have minimized windows, so I don't know if feedreader is running or not. Also, there should be an option for "close feedreader instead of minimizing it".

brendanlong:
> Also, there should be an option for "close feedreader instead of minimizing it".

I would be in favor of making that the only option and not ever running in the background..

Issue #58, opened 2016-02-13
Notifications Icon in Systray


Tuchkata:
Hi, is there an option to add a system tray notification icon? I haven't found one yet.

jangernert:
Nope
Gnome as well as elementary's Pantheon are ditching traditional system trays. Even Ubuntu's "indicators" are no longer supported. So no plans to add this to FeedReader.
There is however support for unity's launcher API to display a badge on the icon in the dock. Ubuntu's Unity, Plank (the dock used by elementary) and some KDE plasmoids support this.

PR #779, opened 2018-10-26
Status icon implementation

naquad:
Hi.

First of all thank you very much for FeedReader. It helped me to improve my workflow a lot: no need to run browser with TTRSS tab open all the time + notifications and awesome UI.

There's one thing that bothered me though: I'm not using Unity and I'm unable to figure out if there are any new items. Not a tragedy, but a bit annoying (checking FeedReader all the time). To cover this case I've implemented a Gtk.StatusIcon support in FeedReader.

Before you hit reject here are my reasons for merging this request:

* It improves UX a lot as you can figure out if FeedReader needs attention at a glance.
* Despite Gtk.StatusIcon is being deprecated there are no other means to provide indicators in Gtk applications. Unity and similar KDE solution are environment specific, they're not available in other window managers. E.g. I'm using Xfce4 and I don't have any other means to provide indication except system tray.
* System tray itself is not going anywhere. It will stay as very many applications still use it. Even if Gtk.StatusIcon will be removed I'm sure some library implementing this functionality will pop up eventually.
* I'm not the only one who would like to have tray icon: #694, #522, #58.

P. S. If it should be implemented in some different way then please let me know and I'll do my best to update the code.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -262,6 +262,11 @@ else
 	message('libUnity support OFF')
 endif
 
+if get_option('statusicon')
+        message('Status icon support ON')
+        vala_args += ['-D', 'STATUSICON']
+endif
+
 # Creating an executable file
 feedreader_lib = shared_library(
   'FeedReader',
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -4,6 +4,11 @@ option('share-plugins',
         value : ['browser', 'email', 'instapaper', 'pocket', 'telegram', 'twitter', 'wallabag'],
         description: 'List of share plugins to install'
 )
+option('statusicon',
+        type: 'boolean',
+        value: false,
+        description: 'Enables status icon'
+)
 option('backend-plugins',
         type: 'array',
         choices: ['bazqux', 'feedbin', 'feedhq', 'feedly', 'fresh', 'inoreader', 'local', 'oldreader', 'owncloud', 'ttrss'],
--- a/src/FeedReader.vala
+++ b/src/FeedReader.vala
@@ -15,12 +15,20 @@
 
 using GLib;
 using Gtk;
+using Gdk;
 
 namespace FeedReader {
 
 	public const string QUICKLIST_ABOUT_STOCK = N_("About FeedReader");
 
 	public class FeedReaderApp : Gtk.Application {
+
+#if STATUSICON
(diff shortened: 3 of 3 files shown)


bilelmoussaoui:
Honestly, I'm against merging this. 
We are using flatpak to ship the latest stable version of FeedReader and it's just a pain to build libappindicator (a lot of patches must be included...) And it doesn't even work with the latest version of the GNOME stack.
The tray icon doesn't work with any HiDPI screen, it doesn't scale up/down. Wil require a lot of work for not that much benefits....

If you have any design ideas to improve the workflow of how we should show that there's some unreaded articles. Please let us know. But I'm just against this.

naquad:
Icon looks fine, at least on 140-160 DPI.
I don't have much other ideas on how to provide indications as except tray icon they all are environment-specific :(

bilelmoussaoui:
I don't have an idea neither. We already use libunity for the systems that supports it. We send notification when something new was fetched, I don't see why we need to add yet an other way of doing the same thing unless we get a new and powerful API to handle those kind of things. The tray icons are not part of Gtk4 and there's no way for us to build a widget like this and stick to the past.

brendanlong:
Does GNOME make it possible to have dynamic app icons in the task switcher?

jangernert:
I'm also a bit hesitant. StatusIcon is deprecated since Gtk+ 3.14, which was released in 2014 according to a quick google search. So already 4 years ago.
Most of the libunity functionality is present in most desktops. Badges are available in Gnome, Pantheon, KDE and Budgie(?). Not 100% sure about budgie to be honest.

brendanlong:
Hm I just installed libunity-devel and rebuilt and nothing seems different to me. Should this do something in GNOME3?

brendanlong:
I'm going to close this because I don't want to leave a pull request hanging when the answer from two of the maintainers is that we don't want to use the system tray. It would be nice to figure out a way to show the number of new items in the GNOME task switcher though.

Thanks for making the pull request and sorry that we won't be merging it :(

jangernert:
@brendanlong by default gnome3 doesn't have any kind of dock or task-bar.
But dash-to-dock for example supports badges
https://micheleg.github.io/dash-to-dock/release/2017/12/25/new-relsease-v62.html

PR #785, opened 2018-10-31
translate zh-TW

doraeric:


bilelmoussaoui:
should be done here instead https://hosted.weblate.org/projects/feedreader/

PR #786, opened 2018-11-05
When a category is removed, uncategorize feeds/categories explicitly to plugins

39aldo39:
Currently, when a category is removed without children, its feeds/categories are uncategorized. However, these changes aren't sent to the plugins. For example, in Nextcloud the children are [removed](https://github.com/nextcloud/news/blob/master/docs/externalapi/External-Api.md#deleting-a-folder).

This pull request sends these changes explicitly to the plugin.

### DIFF
--- a/src/Backend/Backend.vala
+++ b/src/Backend/Backend.vala
@@ -568,6 +568,24 @@ namespace FeedReader {
 
 		public void removeCategory(string catID)
 		{
+			var feeds = DataBase.readOnly().read_feeds();
+			foreach(Feed feed in feeds)
+			{
+				if(feed.hasCat(catID))
+				{
+					moveFeed(feed.getFeedID(), catID);
+				}
+			}
+
+			var cats = DataBase.readOnly().read_categories(feeds);
+			foreach(var cat in cats)
+			{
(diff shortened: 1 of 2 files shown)


brendanlong:
Is it possible that this belongs at the plugin level, in `"deleteCategory"`, and the NextCloud plugin is just implementing that function incorrectly?

jangernert:
I would also suggest to move this into the NextCloud plugin. Most RSS services have children of deleted categories still around as uncategorized.

brendanlong:
Ok I looked at this a little more, and I'm still a little worried because this part of the code is pretty flaky, but I'm pretty sure this pull request won't make it any worse. Hopefully sometime soon we'll be able to test the plugins.

This seems to work in Feedbin (well.. as well as it worked previously, which is "barely"). Do you mind if I merge this @jangernert?

@39aldo39 I took a look at your fork and I'm happy to review a pull reques to add your plugin. I'll try to review it later this week if I can find time, but the major thing that stands out is that we should try not to duplicate the dependencies shared with the local plugin (date parsing, libmrss, etc.).

Issue #730, opened 2018-08-06
Cleaning the database never ends


zeitrechnen:
I installed FeedReader and went through the the setup guide. I added two feeds to my local database and then the app crashed. When I opened it again the message "FeedReader is cleaning the database" appeared. After 30 minutes I quit the app. I removed and reinstalled it. But whenever I open the app FeedReader wants to clean the database and never finishes.

Version: 1.6 from Flathub
OS: Pop!_OS 18.04

bilelmoussaoui:
How did you install 1.6 from flathub??

zeitrechnen:
Yes.

bilelmoussaoui:
That's just impossible. The only version we have pushed to flathub is 2.2

zeitrechnen:
Sorry, is this a different app? https://flathub.org/apps/details/org.gnome.FeedReader

![bildschirmfoto vom 2018-08-06 22-41-08](https://user-images.githubusercontent.com/1907339/43739901-e82170ba-99c9-11e8-8c2c-47427a1e2d99.png)

zeitrechnen:
Weird, seems to be the only app on Flathub with this name. Am I doing something wrong, so that the old version is displayed for me.

bilelmoussaoui:
A new flatpak version should be available soon. Please try that and let us know if you still have this issue

zeitrechnen:
It works now. Thank you!

LBreda:
Same issue on 2.3.0 on flathub

LBreda:
Solved by setting "spring-cleaning" to false on gconf org->gnome->feedreader->feedly

Wdavery:
Still having this issue. It needs to be reopened.

brendanlong:
Reopening since this seems to still be an issue, at least with the Feedly backend.

I've never seen this, but admittedly I clear out of all the app data on a regular basis as part of testing other changes.

@Wdavery  You may want to try our official Flatpak package. We have no control over Solus's packages.

brendanlong:
Do any of you happen to have logs from this? The spring cleaning code is surprisingly simple and I don't really see what could be wrong with it. It literally just runs a `VACUUM` in SQLite :\

It's possible the database is locked or something?

brendanlong:
Ah I think the issue is that we set a persistent variable when it starts and then if it fails we don't retry?

brendanlong:
[...]

PR #788, opened 2018-11-06
Remove spring cleaning from saved state

brendanlong:
The DB vacuum isn't a persistent operation, and it should just restart
next time if it's interrupted. This should fix an issue where if the
spring cleaning was interrupted, it would hang forever.

See #730

### DIFF
--- a/src/Backend/Backend.vala
+++ b/src/Backend/Backend.vala
@@ -25,6 +25,7 @@ namespace FeedReader {
 		private bool m_offline = true;
 		private bool m_cacheSync = false;
 		private uint m_timeout_source_id = 0;
+		private Mutex m_sync_lock;
 		private delegate void asyncPayload();
 
 		public signal void syncStarted();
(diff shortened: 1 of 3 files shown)


src/Backend/Backend.vala
```diff
@@ -185,62 +186,69 @@ namespace FeedReader {
 
 		private void sync(bool initSync = false, GLib.Cancellable? cancellable = null)
 		{
-			if(Settings.state().get_boolean("currently-updating")
-			|| Settings.state().get_boolean("spring-cleaning"))
+			// Prevent multiple concurrent syncs or spring cleanings
+			// We would prefer to use MutexLocker but it doesn't seem to work
+			m_sync_lock.lock();
```
brendanlong: Using a mutex here instead of tricky state variables

brendanlong:
To test this, I manually set /org/gnome/feedreader/saved-state/spring-cleaning to true in dconf-editor, opened feedreader, and confirmed that it was hanging claiming to do a spring cleaning.

Then I applied this patch and opened it again and it loaded fine. Running and sync and spring cleaning was instant. I also repeated the dconf change and it worked fine a second time.

PR #793, opened 2018-11-22
Various improvements

brendanlong:


jangernert:
Nice, didn't know about these kind of constraints on function arguments. Just curious what happens if the conditions aren't met? Is an error thrown that can be caught on a higher level? Google didn't have an answer for me there.

brendanlong:
> Nice, didn't know about these kind of constraints on function arguments. Just curious what happens if the conditions aren't met? Is an error thrown that can be caught on a higher level? Google didn't have an answer for me there.

They're called "contracts" and basically get translated into assertions:

https://wiki.gnome.org/Projects/Vala/Tutorial#Assertions_and_Contract_Programming

I figured it would be nice to find out about bad state through stuff like this rather than segfaults, although it's a little tricky since you have to make sure the assertions actually always pass (and sometimes it's hard to tell what values are actually allowed). For now I'm being fairly conservative about them.

PR #797, opened 2018-11-24
Use CircleCI workflows

brendanlong:
This makes tests separate from builds. Yay I guess?


brendanlong:
Hm I can't figure out how to make this work.

PR #801, opened 2018-11-25
Copy article url from ArticleRow

Benguin:
Some small changes to allow quickly copying the article URL for sharing.


jangernert:
Seems fine. Thx :)

PR #803, opened 2018-11-26
Change Free Software icon to Open Source

### LINKED ISSUE #794: Feedbin isn't proprietary or GPL
I just noticed that the Feedbin plugin marks it as proprietary:

```
	public BackendFlags getFlags()
	{
		return (BackendFlags.HOSTED | BackendFlags.PROPRIETARY | BackendFlags.PAID);
	}
```

But the source code of Feedbin is released under the MIT license, which qualifies as Free Software. I'm not sure if I should switch to BackendFlags.FREE_SOFTWARE though since that causes us to show a "GPL" logo, and it's not GPL.

Any idea what we should do here?

brendanlong:
This changes the free software/GPL option to Open Source so we can
use an icon that doesn't imply a particular license.

It also changes the Feedbin plugin to be listed as open source.

Fixes #794


brendanlong:
I created this pull request on OSX, so I haven't built/tested yet. I'll make sure it looks good on Linux before merging.

Need a review from @jangernert  though to make sure this change is ok.

jangernert:
The copyright logo in the bottom right of the icon doesn't really make sense, since the icon will be scaled down significantly leaving the copy right logo as just a few random pixels.
Also symbolic icons should be 1 solid color. Not sure if the gray tone used in the other icons was necessary or just recommended

brendanlong:
Hmm ok. I might need to think about this more. The OSI logo's copyright say you have to keep the stupid (R). I'm not sure if we're allowed to make it a single color.

brendanlong:
I think I'm just going to set Feedbin to `FREE_SOFTWARE` and deal with the confusing icon later.

PR #804, opened 2018-11-27
Add DecSync plugin

39aldo39:
This pull request adds a DecSync plugin for synchronization using [DecSync](https://github.com/39aldo39/DecSync). It is based on the local plugin, with some syncing added.

It also adds the possibility for plugins to always use the articleIDs when articles are read, instead of the `setFeedRead`, `setCategoryRead` and `markAllItemsRead` methods. This is needed since DecSync only uses the articleIDs. Moreover, it is not possible to get the IDs from the database, since it can be updated before the plugin method is called. This also may be more correct for other plugins, since there can be new articles between the last sync and the read action, especially when cached actions are used.

Another point as discussed [here](https://github.com/jangernert/FeedReader/pull/786#issuecomment-441334721) is that this duplicates some dependencies from the local plugin. Probably the best way is to move them out of the plugin directories.


plugins/backend/decsync/decsyncUtils.vala
```diff
@@ -0,0 +1,120 @@
...
+			return text;
+
+		try
+		{
+			return GLib.convert(text, -1, "utf-8", locale);
+		}
+		catch(ConvertError e)
+		{
+			Logger.error(e.message);
+		}
```
brendanlong: It's not clear to me if catching this error is a good idea, since from the user's perspective we'll be silently ignoring whatever text we tried to convert.

plugins/backend/decsync/decsyncListeners.vala
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	private static void addCategory(decsyncInterface plugin, string catID)
+	{
+		if (catID == plugin.uncategorizedID() || catID == CategoryID.MASTER.to_string() || plugin.m_db.read_category(catID) != null)
+		{
+			return;
+		}
+		var cat = new Category(catID, catID, 0, 99, CategoryID.MASTER.to_string(), 1);
+		var list = new Gee.LinkedList<Category>();
```
brendanlong: You can use `ListUtils.single(cat)` to generate a list with one element as efficiently as possible.

(Also, LinkedLists are almost always worse than ArrayLists)

plugins/backend/decsync/decsyncInterface.vala
```diff
@@ -0,0 +1,738 @@
...
+	}
+
+	public string getServerURL()
+	{
+		return "http://localhost/";
+	}
+
+	public string uncategorizedID()
+	{
+		return "0";
```
brendanlong: Should this return `CategoryID.MASTER.to_string()`?

plugins/backend/decsync/decsyncInterface.vala
```diff
@@ -0,0 +1,738 @@
...
+
+	public string createTag(string caption)
+	{
+		string tagID = "1";
+
+		if(!m_db.isTableEmpty("tags"))
+			tagID = (int.parse(m_db.getMaxID("tags", "tagID")) + 1).to_string();
+
+		Logger.info("createTag: ID = " + tagID);
+		return tagID;
```
brendanlong: I'm confused about this function since `supportTags()` is false. This function should probably either return an empty string or throw an exception.

(At some point I want to redo how plugins work so these functions don't even exist unless the plugin advertises supporting them, but for now just not having any logic in them is good enough)

plugins/backend/decsync/decsyncInterface.vala
```diff
@@ -0,0 +1,738 @@
...
+		feedID = feedURL;
+
+		Logger.info(@"addFeed: ID = $feedID");
+		Feed? Feed = m_utils.downloadFeed(m_session, feedURL, feedID, catIDs, out errmsg);
+
+		if(Feed != null)
+		{
+			if(!m_db.feed_exists(Feed.getURL())) {
+				var list = new Gee.ArrayList<Feed>();
+				list.add(Feed);
```
brendanlong: Consider using `ListUtils.single(feed)`

plugins/backend/decsync/decsyncInterface.vala
```diff
@@ -0,0 +1,738 @@
...
+		}
+		else
+		{
+			catIDs.add(uncategorizedID());
+		}
+
+		feedID = feedURL;
+
+		Logger.info(@"addFeed: ID = $feedID");
+		Feed? Feed = m_utils.downloadFeed(m_session, feedURL, feedID, catIDs, out errmsg);
```
brendanlong: Variables should be lowercase:

```
Feed? feed = m_utils.downloadFeed(m_session, feedURL, feedID, catIDs, out errmsg);
```

brendanlong:
This seems to be causing a build failure in CircleCI due to a missing meson.build file:

https://circleci.com/gh/jangernert/FeedReader/632

Also, is any of this easily testable? I'm trying to get more pieces of FeedReader unit-tested, although I admit I don't know how to do async tests yet. If you think any of it can be tested, you can see an example in the Feedbin plugin (TestFeedbin.vala and the last section of the feedbin plugin's meson.build).

39aldo39:
I have implemented most of requested changes. Only the function from `localUtils.downloadFeed` still uses error messages. Some changes applied to the local backend as well. It is a bit ugly that some code is almost the same as the local backend. But I don't think it is easy to deduplicate it, since there are still some small changes.

The CircleCI build fails since `libdecsync` is a submodule, which probably isn't cloned.

I do think this plugin is testable, as long writing/reading files is possible. We can then test whether the changes are correctly written/executed.

brendanlong:
> It is a bit ugly that some code is almost the same as the local backend. But I don't think it is easy to deduplicate it, since there are still some small changes.

Yeah that's fine. The plugins probably shouldn't depend on each other, I just didn't want to libraries they use to be duplicated too.

> The CircleCI build fails since libdecsync is a submodule, which probably isn't cloned.

I just created #809 to fix this. If you rebase once that lands on master it should fix it.

I'll review the updated code after work today.

plugins/backend/local/localInterface.vala
```diff
@@ -393,9 +393,7 @@ public class FeedReader.localInterface : Peas.ExtensionBase, FeedServerInterface
 		{
 			string cID = createCategory(newCatName, null);
 			var cat = new Category(cID, newCatName, 0, 99, CategoryID.MASTER.to_string(), 1);
-			var list = new Gee.ArrayList<Category>();
-			list.add(cat);
-			m_db_write.write_categories(list);
+			m_db_write.write_categories(ListUtils.single(cat));
```
brendanlong: Thanks for fixing these too! I didn't realize this was existing code you had copied.

brendanlong:
I just got a chance to test this and it seems to work. Thanks for creating this!

brendanlong:
I just triggered a v2.6.0 release so this should show up in Flathub soon.

PR #811, opened 2018-11-29
Add a Makefile

brendanlong:
This is just a helper so I can type `make` instead of
`meson builddir && ninja -C builddir`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,17 @@
+BUILDDIR := builddir
+
+default: build
+
+$(BUILDDIR):
+	meson $(BUILDDIR)
+
+build: $(BUILDDIR)
+	ninja -C $(BUILDDIR)
+
+test: $(BUILDDIR)
+	ninja -C $(BUILDDIR) test
+
+install: build
+	ninja -C $(BUILDDIR) install
+
+.PHONY: build test install default


brendanlong:
These passthrough makefiles are pretty common in OCaml but I don't know if this is too weird in meson-land?

jtojnar:
When GNOME started switching to Meson, they adopted common [`build-api`](https://github.com/cgwalters/build-api) convention. But they started removing the Makefiles again, shortly after their continuous integration system gained support for Meson.

The wrappers are usually too limited to be of any non-trivial use anyway; often, they fail at controlling features like cross-compilation or parallelization. Furthermore, in my opinion, they do not offer any benefit to specifying the build instructions in `README`, as we are [already doing](https://github.com/jangernert/FeedReader#build), since even with `Makefile`, you still need to check the `README` or `Makefile` to see which convention it follows.

brendanlong:
I was thinking of using a Makefile basically for developer convenience (`make` and `make test` are less annoying to type). I was assuming packagers would still use meson/ninja directly, but maybe adding a Makefile would be confusing?

jtojnar:
I think that is what shell aliases are for.

brendanlong:
I decided I wanted a Makefile anyway and added it in 88cf9b92db44923f5c9cc33a58fc4597d8581cc7

It's not strictly necessary but there's something to be said for making the developer experience not annoying.

Issue #814, opened 2018-12-04
plugins/backend/decsync/ is empty


raczkow:
When you try to download zip/tar.gz archive with feedreader 2.6.0 (from https://github.com/jangernert/FeedReader/releases and from https://github.com/jangernert/FeedReader/archive/master.zip), you get an empty directory plugins/backend/decsync/ which leads to compilation errors:

```
plugins/backend/decsync/meson.build:6:0: ERROR:  Non-existent build file 'plugins/backend/decsync/libdecsync/src/meson.build'
```

brendanlong:
Duplicate of #813 

I'm going to undo this release until I figure out how to make it work.

brendanlong:
Actually not exactly a duplicate.

brendanlong:
I think our options here are:

 1. Make custom release artifacts that have the submodule code
 2. Inline the submodule, either just by literally copying it or by using a subtree
 3. Remove the submodule and build/install the dependency instead of vendoring it

I'm leaning toward option 2

brendanlong:
I just did a new v2.6.1 release which merges all of this stuff into our repo. Can you take a look and see if that works for you?

raczkow:
It's OK now.

Issue #813, opened 2018-12-04
Handle submodules in Flatpak


brendanlong:
The latest release isn't working because it needs a git submodule to load and I'm not sure how to do that in Flatpak. Maybe we should just add it as a separate source but I'm not sure how to link it into the build directory.

I wonder if this also caused problems with grabber configs since those are also a submodule?

brendanlong:
You can install libdecsync-vala using a module like this:

```

    {
      "name": "libdecsync-vala",
      "buildsystem": "meson",
      "sources": [{
        "type": "git",
        "url": "https://github.com/39aldo39/libdecsync-vala.git",
        "commit": "30681106cf38c4017e428630ee365e1a15eed318"
      }]
    },
```

I don't know how to get that copied into the build directory for feedreader though.

bilelmoussaoui:
Will take a look

petefwalter:
Same thing here: can't build the fedora package any more due to this.

brendanlong:
I just did a new v2.6.1 release which merges all of this stuff into our repo. Can you take a look and see if that works for you?

petefwalter:
It does. Thanks!

PR #816, opened 2018-12-07
Subtrees

brendanlong:
This switches our submodules into git subtrees:

https://www.atlassian.com/blog/git/alternatives-to-git-submodule-git-subtree

The main advantage of subtrees is that the code is in our repo, so our releases
will have all of the expected data.

For convenience, I also added a new script to pull changes to these:

```bash
./subtree_pull.sh libdecsync
./subtree_pull.sh grabberconfig
```

This also updates us to the latest version of the grabber configs because it was
easier than loading the old version we had before.

See #814 and #813


brendanlong:
Merged with 3cc7df1812ddbd484ac1493d178d1ad58b389d4b

PR #820, opened 2018-12-20
Add access to the default DecSync directory (flatpak)

39aldo39:


org.gnome.FeedReader.json
```diff
@@ -22,7 +22,10 @@
     "--talk-name=org.gnome.OnlineAccounts",
     "--own-name=org.gnome.FeedReader.ArticleView",
     "--talk-name=org.freedesktop.Notifications",
-    "--talk-name=org.freedesktop.secrets"
+    "--talk-name=org.freedesktop.secrets",
+    /* Access to DecSync directory */
+    "--env=DECSYNC_DIR=.local/share/decsync",
+    "--filesystem=~/.local/share/decsync"
```
bilelmoussaoui: Why is this needed? I have no idea how desync works that's why I'm asking

39aldo39: DecSync uses a directory on the filesystem to read/write updates about e.g. read articles. This directory then is synced with other devices.

As for the environment variable, flatpak sets `XDG_DATA_HOME` to some obscure path, so I have to create a new one to get a reasonable default.

bilelmoussaoui: Alright, so the Desync plugin writes those data on that directory and DeSync can pull the data from there to sync it with other devices?

bilelmoussaoui:
This seems pretty okay to me. Even If having some kind of portal to access the Desync data would have been a much better solution as currently every app can read, write and open those files. 

Merged.

bilelmoussaoui:
Could you open a PR on the Flathub repository with those changes too? Thanks!

PR #822, opened 2018-12-24
Don't use legacy path for AppStream metainfo file

City-busz:
Metainfo files should be installed into /usr/share/metainfo.


bilelmoussaoui:
Thanks!

PR #825, opened 2019-01-03
Fix crashes caused by ArticleList.m_loadThread

39aldo39:
Fix some crashes caused by `ArticleList.m_loadThread` being used after being joined. Only the last join in `updateArticleList` is a problem, as it isn't initialized after the join. But by setting `m_loadThread` explicitly to `null` after each join the problem is clearly fixed.


brendanlong:
Thanks!

Issue #826, opened 2019-01-05
Crash in random times


gavr123456789:
I can not understand the reason for the departure, but the application does not work longer than 3 minutes.
Logs:
https://pastebin.com/er4mFgUF

brendanlong:
I made a major change in #843 to try to address crashes like this. Can you check if you get this error in the latest (v2.7.0)?

brendanlong:
I think this was caused by the shared non-threadsafe database connections that were fixed in #843. Please let us know if you see any other problems with the latest release.

PR #828, opened 2019-01-07
Add mutex to database transactions

39aldo39:
This PR adds a mutex to database transactions to prevent nested transactions in concurrent writes. Maybe there should be more locks on writing to the database, but I think locking just transactions should fix most problems. It is also simple and should not cause other problems.

### DIFF
--- a/src/DataBaseWriteAccess.vala
+++ b/src/DataBaseWriteAccess.vala
@@ -16,6 +16,7 @@
 public class FeedReader.DataBase : DataBaseReadOnly {
 
 private static DataBase? m_dataBase = null;
+private static Mutex m_transactionLock;
 
 public static new DataBase writeAccess()
 ensures (m_dataBase != null)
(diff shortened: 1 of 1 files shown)


brendanlong:
We shouldn't be using database connections across threads anymore after 13e745bcf675aa8a5e9bf28996477b5a8e7eaf0d. Do you know where the problems this fixes are coming from? Is one of the plugins still using the DB concurrently?

39aldo39:
I sometimes get crashes with the local and decsync plugins. It happens while adding the initial feeds (either the suggested ones for the local plugin or the stored ones for the decsync plugin). See also the log in #826. The crashes occur in `DataBaseWriteAccess.write_feeds` or `DataBaseWriteAccess.write_categories`.

brendanlong:
Thanks for the patch. I think you're probably right about the cause of the problem but I think our database/threading issues were serious enough that this wouldn't fix it. I just landed #843 which should ensure that we never use database connections in unsafe ways. I'll do a new release now. Can you let me know if it works for you (it will be version 2.7.0)

PR #842, opened 2019-01-28
meson: fix finding libcurl

worldofpeace:
`compiler.find_library` expects the `lib_name` without `lib` prefix.


brendanlong:
Thanks for the patch! Is this causing problems for you somewhere or is this just a build cleanup?

I read some of the related documentation and it seems like maybe `c_compiler.find_library` isn't the right thing to use here anyway, since we should use `dependency`. I'm not very familiar with Meson still so do you know if this is right?

brendanlong:
Ahh nevermind I see, if I we use `dependency` it tries to find a Vala library named `libcurl`.

brendanlong:
Thanks again for the patch!

worldofpeace:
Yeah this was causing problems for us in NixOS https://github.com/NixOS/nixpkgs/commit/2c84b7b4a140394198824b5936b68d7de9e9eeb4

Thanks for the quick merge :smile:

brendanlong:
Ah makes sense

Issue #836, opened 2019-01-18
Crash after adding a new feed


Strubbl:
OS: Archlinux
WM: i3wm

$ feedreader --version
Version: 2.6.0
Git Commit: 939ea58

When i add a new feed, the app crashes with the following error:

```
An error occured (SIGSEGV)
   The reason is likely a null reference being used.
   Note: no file path and line numbers can be retrieved. Are you sure you added -g -X -rdynamic to valac command line?

    #1  <unknown>   in 'g_thread_join'
        at /usr/lib/libglib-2.0.so.0
    #2  <unknown>   in 'feed_reader_article_list_updateArticleList'
        at /usr/lib/libFeedReader.so
    #3  <unknown>   in 'g_main_context_dispatch'
        at /usr/lib/libglib-2.0.so.0
    #4  <unknown>   in 'g_main_context_iteration'
        at /usr/lib/libglib-2.0.so.0
    #5  <unknown>   in 'g_application_run'
        at /usr/lib/libgio-2.0.so.0
    #6  <unknown>   in 'feed_reader_main_main'
        at feedreader
    #7  <unknown>   in '__libc_start_main'
        at 
    #8  <unknown>   in '_start'
        at feedreader
```

brendanlong:
I made a major change in #843 to try to address crashes like this. Can you check if you get this error in the latest (v2.7.0)?

brendanlong:
The `thread.join()` that was crashing doesn't exist anymore, so this should be fixed.

Issue #834, opened 2019-01-18
Crash when scrolling to the bottom of the article list


crackwilding:
Scroll to the bottom of the article list and (most of the time) the application quits. FeedReader 2.6.0 on Pop OS 18.10 / Gnome 3.30.2. I tailed the syslog and fiddled with it until it crashed. Results below:

Jan 18 09:58:09 elwood org.gnome.FeedReader.desktop[2203]: (feedreader:2): GdkPixbuf-DEBUG: 09:58:09.290: gdk_pixbuf_from_pixdata() called on:
Jan 18 09:58:09 elwood org.gnome.FeedReader.desktop[2203]: (feedreader:2): GdkPixbuf-DEBUG: 09:58:09.291: #011Encoding raw
Jan 18 09:58:09 elwood org.gnome.FeedReader.desktop[2203]: (feedreader:2): GdkPixbuf-DEBUG: 09:58:09.291: #011Dimensions: 14 x 14
Jan 18 09:58:09 elwood org.gnome.FeedReader.desktop[2203]: (feedreader:2): GdkPixbuf-DEBUG: 09:58:09.291: #011Rowstride: 56, Length: 808
Jan 18 09:58:09 elwood org.gnome.FeedReader.desktop[2203]: (feedreader:2): GdkPixbuf-DEBUG: 09:58:09.291: #011Copy pixels == false
Jan 18 09:58:09 elwood org.gnome.FeedReader.desktop[2203]: (feedreader:2): GdkPixbuf-DEBUG: 09:58:09.348: gdk_pixbuf_from_pixdata() called on:
Jan 18 09:58:09 elwood org.gnome.FeedReader.desktop[2203]: (feedreader:2): GdkPixbuf-DEBUG: 09:58:09.348: #011Encoding raw
Jan 18 09:58:09 elwood org.gnome.FeedReader.desktop[2203]: (feedreader:2): GdkPixbuf-DEBUG: 09:58:09.348: #011Dimensions: 14 x 14
Jan 18 09:58:09 elwood org.gnome.FeedReader.desktop[2203]: (feedreader:2): GdkPixbuf-DEBUG: 09:58:09.348: #011Rowstride: 56, Length: 808
Jan 18 09:58:09 elwood org.gnome.FeedReader.desktop[2203]: (feedreader:2): GdkPixbuf-DEBUG: 09:58:09.348: #011Copy pixels == false
[...]

PR #843, opened 2019-01-28
Remove threading

brendanlong:
There are a bunch of errors related threading and unsafe sharing of databases and possibly other data structures. This pull request just strips all of that out, and makes every database handle only be used in one place and then released. This is not particularly efficient, but the UI is still responsive and I don't have time to fix the old threading situation. Working correctly seems better than being perfectly efficient.

See #840, #836, #834, #828 and possibly others.

### DIFF
--- a/plugins/backend/bazqux/bazquxAPI.vala
+++ b/plugins/backend/bazqux/bazquxAPI.vala
@@ -162,6 +162,7 @@ public bool getCategoriesAndTags(Gee.List<Feed> feeds, Gee.List<Category> catego
 	uint length = array.get_length();
 	int orderID = 0;
 
+	var db = DataBase.readOnly();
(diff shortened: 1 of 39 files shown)


brendanlong:
For future reference, the problems we have are that we did a bunch of async stuff using background threads, but most of FeedReader isn't actually threadsafe. If we want to do this, we'll need to wrap each thing we use to be async-friendly (like our DB code) and then use `async` instead of threads.

brendanlong:
The one thing I'm worried about is that I'm using a fast SSD. It would be nice if I could get someone with a spinning-disk to test this out and see if the SQLite change is ok.

brendanlong:
Ah I might be able to test this using an external hard drive..

PR #851, opened 2019-02-08
Improve error messages when adding a feed fails

brendanlong:
Previous a lot of errors just said "Failed to add the feed", which was confusing
for users.

 - Always show the URL we attempted to download in the error message
 - Add a special message for network errors
 - Show the HTTP status code if we got an error
 - Explain more specifically what happened when parsing fails

### DIFF
--- a/plugins/backend/decsync/decsyncUtils.vala
+++ b/plugins/backend/decsync/decsyncUtils.vala
@@ -43,23 +43,34 @@ public void setDecsyncDir(string decsyncDir)
 	Utils.gsettingWriteString(m_settings, "decsync-dir", decsyncDir);
 }
 
-public Feed? downloadFeed(Soup.Session session, string xmlURL, string feedID, Gee.List<string> catIDs, out string errmsg)
+public Feed? downloadFeed(Soup.Session session, string feed_url, string feedID, Gee.List<string> catIDs, out string errmsg)
 {
-	errmsg = "";
+	var error = new StringBuilder(_("Failed to add feed"));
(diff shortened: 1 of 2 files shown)


brendanlong:
screenshot from 2019-02-08 18-02-43
screenshot from 2019-02-08 18-00-12
screenshot from 2019-02-08 18-00-01

plugins/backend/decsync/decsyncUtils.vala
```diff
@@ -43,23 +43,27 @@ public void setDecsyncDir(string decsyncDir)
 	Utils.gsettingWriteString(m_settings, "decsync-dir", decsyncDir);
 }
 
-public Feed? downloadFeed(Soup.Session session, string xmlURL, string feedID, Gee.List<string> catIDs, out string errmsg)
+public Feed? downloadFeed(Soup.Session session, string feed_url, string feedID, Gee.List<string> catIDs, out string errmsg)
 {
 	errmsg = "";
+	string error_prefix = _(@"Failed to add feed $feed_url\n");
```
brendanlong: @bilelmoussaoui Is this the right way to make strings translatable? Do I need to do anything else, and is it ok that it has a variable in it?

bilelmoussaoui: You should avoid having variables there and use printf, otherwise translators might translate the variable name too.

brendanlong: Thanks, I switched this so the translatable strings don't have variables in them.

PR #852, opened 2019-02-09
TT-RSS improvements

brendanlong:
Various improvements to clean up the TT-RSS plugin, mostly involving stronger typing.


brendanlong:
I need to find someone who's running TT-RSS so they can test this. I haven't tested this at all besides the fact that it compiles.

brendanlong:
Ok, I did some local testing and I'm pretty sure everything works now, although the feed-adding logic is still not quite right, but it does eventually work.

I only tested this with Postgres but the dynamically typed JSON stuff should make it work in MySQL too.

PR #854, opened 2019-02-09
Fix #853: use dc:creator tag for author, author tag for author email

### LINKED ISSUE #853: RSS 'author' is an email. 'dc:creator' is the author name
If "Local RSS" is selected, and you add an RSS feed, the author of the post will not be visible in the header next to the date.

This is because the "author" tag is being read for the post author in the RSS parser, but actually, "author" is the tag for the author's email. The correct tag for the author is "dc:creator:" See:
http://www.lowter.com/blogs/2008/2/9/rss-dccreator-author
https://cyber.harvard.edu/rss/rss.html#hrelementsOfLtitemgt

I have a PR to fix this which I can submit, in the libmrss dependency. But I wonder why this library is included in the source -- should this be fixed upstream instead (e.g. at https://www.autistici.org/bakunin/libmrss/doc/)?

sdjacobs:
Fixes libmrss library to read correct tag for a post's author.

(Perhaps this should be fixed upstream instead of in this repository.)

### DIFF
--- a/libraries/libmrss/mrss_parser.c
+++ b/libraries/libmrss/mrss_parser.c
@@ -720,8 +720,13 @@ __mrss_parser_rss_item (nxml_t * doc, nxml_data_t * cur, mrss_t * data)
 		}
 	    }
 
+	  /* author email */
+	  else if (!strcmp (cur->value, "author") && !item->author_email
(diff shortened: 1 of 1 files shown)


brendanlong:
Thanks!

PR #859, opened 2019-02-10
primary menu: follow gnome design team guidelines

bilelmoussaoui:
https://gitlab.gnome.org/GNOME/Initiatives/issues/4

### DIFF
--- a/src/Utils.vala
+++ b/src/Utils.vala
@@ -752,17 +752,13 @@ public static GLib.Menu getMenu()
 	var aboutMenu = new GLib.Menu();
 	aboutMenu.append(Menu.shortcuts, "win.shortcuts");
 	aboutMenu.append(Menu.about, "win.about");
-	aboutMenu.append(Menu.quit, "app.quit");
 
 	var menu = new GLib.Menu();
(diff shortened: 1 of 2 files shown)


src/Utils.vala
```diff
@@ -752,17 +752,13 @@ public static GLib.Menu getMenu()
...
 	aboutMenu.append(Menu.shortcuts, "win.shortcuts");
 	aboutMenu.append(Menu.about, "win.about");
-	aboutMenu.append(Menu.quit, "app.quit");
 
 	var menu = new GLib.Menu();
 	menu.append(Menu.settings, "win.settings");
 	menu.append(Menu.reset, "win.reset");
-	menu.append_section("", urlMenu);
 
-	if(GLib.Environment.get_variable("XDG_CURRENT_DESKTOP").down() != "pantheon")
```
bilelmoussaoui: This was added here mostly because elementary removed about dialog from all it's core apps. We don't really follow elementary design guidelines and the app is not distributed for their users anymore. That's the reason of the removal of this condition, we should avoid bloating up the whole code with that

PR #877, opened 2019-02-13
Replace Droid fonts with Noto fonts

tim77:
Droid fonts being superseded by Noto fonts
https://github.com/googlei18n/noto-fonts/issues/555

It looks better and more readable especially with mixed languages also it well supported by any distro.

## Droid:
droid

## Noto:
noto

### DIFF
--- a/schemas/org.gnome.feedreader.gschema.xml
+++ b/schemas/org.gnome.feedreader.gschema.xml
@@ -69,7 +69,7 @@
 		</key>
 
 		<key name='font' type="s">
-			<default>'Droid Sans 12'</default>
+			<default>'Noto Sans 12'</default>
 		</key>
 
 		<key name='drop-articles-after' enum='org.gnome.feedreader.drop-articles-duration'>


brendanlong:
I'm largely in favor of this, but I'm not sure if the major distros ship Noto fonts by default. It doesn't seem like Fedora does. I wonder if we can package these into the Flatpak?

brendanlong:
It's actually odd that we set this at all. Why don't we use the GNOME default "document text" font, at least as the default?

tim77:
> I'm largely in favor of this, but I'm not sure if the major distros ship Noto fonts by default. It doesn't seem like Fedora does. I wonder if we can package these into the Flatpak?

Fedora by default ship only Noto Chinese, Japanese, and Korean versions, but always possible to add `Recommends: noto` to spec file. Not sure about Flatpak and policy bundling fonts.

tim77:
> 
> 
> It's actually odd that we set this at all. Why don't we use the GNOME default "document text" font, at least as the default?

Yes, that what i thought when first time noticed that. :) Maybe better just set default GNOME font and thats it.

brendanlong:
I just created #881 to use the system document font as the default so we don't need to think about what fonts are installed.

brendanlong:
I just merged #881 to handle this. Thanks for the info and pull request!

PR #881, opened 2019-02-16
Change the default font to use the system "document font"

brendanlong:
Previously we defaulted to Droid fonts, but we shouldn't be trying to
make assumptions about what fonts users have installed or what they
want to use here.

This makes it so we use the system document font by default and we
still allow users to override it at-will in the settings.

See #877


brendanlong:
Note: Changing the type of this setting will cause it to be reset when people update to this version of FeedReader. Another option would be to leave "font" as a string but default to "". I like using an explict maybe type here better but I could be persuaded that this will annoy enough people that we should keep this as a string.

PR #886, opened 2019-02-20
Assume XML docs are UTF-8 in libnxml

### LINKED ISSUE #476: Unicode character problem, themes, fonts
I've tried FeedReader with my hosted Tiny Tiny RSS but  I found the font problem with Unicode characters in the title and item list.
screenshot_2017-05-09_15-42-05
It's still displayed correctly in item's content.
I looked around for the fonts and  article view settings but I'm not found any information about them.
I guest the problem is the font wasn't support Unicode characters as well. How can I change it and how can I edit the article view theme.

brendanlong:
Previously this required XML documents to explicitly list themselves as
UTF-8 for us to handle UTF-8 codepoints correctly. This switches that
to assume we're using UTF-8 unless the charset is explicitly set.

Fixes #476

### DIFF
--- a/libraries/libnxml/nxml_utf.c
+++ b/libraries/libnxml/nxml_utf.c
@@ -442,7 +442,7 @@ int __nxml_utf_detection(char *r_buffer, size_t r_size, char **buffer,
 int64_t
 __nxml_int_charset(int ch, unsigned char *str, char *charset)
 {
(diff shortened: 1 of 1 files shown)


brendanlong:
Example from <http://alt1040.com/feed>:

screenshot from 2019-02-20 13-38-21

Issue #864, opened 2019-02-11
Switching from All / Unread lose track of the current article


hfiguiere:
If I switch to All / Unread, the current article becomes unselected, losing the position.

jangernert:
If the selected article is also present in the new article-list it will be re-selected.
If it is not in the new list it doesn't make sense to me to still have the article showing in the article-view.
So I'd say `wontfix` unless I misunderstood the report.

hfiguiere:
Switching from 'Unread' to 'All', so the article should still be in the list.

Even the other way I expect it, because it is selected.

brendanlong:
Hmm I'll try to take a look at this sometime. Which backend are you using (local, Feedbin, etc.)?

hfiguiere:
Local.

brendanlong:
I'm largely in favor of leaving the currently selected article open even if it's not on the new list, but I'm not sure about it from a UI perspective, since I feel like there should be a way to close the current article. Any ideas?

hfiguiere:
It could just disappear when a different selection occurs, like moving to the next or previous article.

brendanlong:
That would mean that once you select an article, there's no way to ever close it without opening a different article. Maybe I'm weird but I like to be able to close articles when I'm done reading :\

weierophinney:
@brendanlong I agree with @hfiguiere .

I just started using FeedReader a few weeks ago, and have it using my Nextcloud instance as the backend. If I'm reading an article and a sync operation occurs... the entire UI reloads, and the article disappears **as I'm reading it**. This is incredibly frustrating when it happens, and it's happened to me three times today alone.

From the UI perspective, if the article is currently being viewed, the article view should not be closed. Once you navigate to another article or feed, it can disappear, but not before.

Alternately, add a "close" button to the article view so users can close it without navigating. Once that happens, again, it could disappear from the feed view.

I'd be happy to test any solutions you come up with.

brendanlong:
I agree that we shuld do something like this. I don't have time to work on it right now though.

brendanlong:
This should be somewhat better in v2.9.0 since we restore the article state in more cases, but the full fix will be in v2.10.0 (see #907)

PR #887, opened 2019-02-20
Keep track of currently selected article

brendanlong:
This lets us restore it when people go from one list to another, when
the article is still visible on the new list.

See #864


brendanlong:
Ideally we wouldn't refresh the article view when this happens but that's a lot more complicated to fix.

PR #902, opened 2019-05-13
Use "oldest article first" sorting switch everywhere

### LINKED ISSUE #901: Article Oldest First sorting doesn't work
In Settings, under Article List, when I change the "Oldest First" switch, nothing happens at all.
Articles are sorted newest-first in my window, and I would like to have them oldest-first.

brendanlong:
This was previously restricted to just the Unread articles tab, but
the hover text for this was non-obvious and it seems strange to have
this setting only apply on one tab.

Fixes #901

### DIFF
--- a/src/DataBaseReadOnly.vala
+++ b/src/DataBaseReadOnly.vala
@@ -507,7 +507,7 @@ public class FeedReader.DataBaseReadOnly : GLib.Object {
 
 				query.select_field(order_by);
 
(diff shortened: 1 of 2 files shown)


jangernert:
I'm pretty sure we limited the behavior to the unread column to avoid some weird interactions / bugs. Not sure what exactly it was however.

brendanlong:
We've fixed a bunch of weird interaction logic so let's see what happens if we merge this..

PR #906, opened 2019-05-24
Close article button

brendanlong:
See #864

After this,  I plan to make it so switching articles doesn't close the current article, even if the current article isn't in the new list.


brendanlong:
Screenshot from 2019-05-24 16-58-59
Screenshot from 2019-05-24 16-58-50

brendanlong:
What do you think of this @bilelmoussaoui

jangernert:
The two `X` icons in the same headerbar look a bit weird. But I can't think of a better alternative icon.

PR #922, opened 2019-07-13
Save TT-RSS HTTP auth info

### LINKED ISSUE #917: tt-rss protected with basic auth does not work
Hi,

Connecting to a tt-rss instance protected by basic auth does not seem to work.

Feedreader does not allow the couple username + password to be empty, which is however possible in tt-rss if you use external authentication from basic auth.

In addition, it doesn't seem to send the basic auth credentials it has asked after the first authentication attempt failed. I keep getting `"POST /api/ HTTP/1.1" 401 932 "-" "FeedReader 2.9.2"` in Apache logs.

brendanlong:
Previously we had an auth input but we ignored it.

Fixes #917


brendanlong:
I tested this on my machine and was able to log into a TT-RSS instance using HTTP auth + TT-RSS auth.

PR #923, opened 2019-07-13
HTMLClean improvements

### LINKED ISSUE #918: Mis-rendering of feed titles containing HTML
Description: https://utcc.utoronto.ca/~cks/space/blog/web/AtomTitlesAndFeedReaders

Example blog post: https://utcc.utoronto.ca/~cks/space/blog/web/PreDoesVeryLittle

This should be rendered as `The HTML <pre> element doesn't do very much` but FeedReader renders it as `The HTML element doesn't do very much`. I can confirm that in Feedbin (my backend), it gets rendered correctly:

Screenshot from 2019-06-17 16-14-51

So the issue is definitely on our side:

Screenshot from 2019-06-17 16-15-18

This may be tricky for the local plugin because I think some feeds include HTML incorrectly, but I think it's more important to handle correct feeds than incorrect, and in the case of non-local backends, we should unconditionally trust the server to handle this.

brendanlong:
- Add tests
- TODO: fix #918

### DIFF
--- a/libraries/htmlclean/htmlclean.c
+++ b/libraries/htmlclean/htmlclean.c
@@ -26,6 +26,37 @@
 #include "glib.h"
 #include "gumbo.h"
 
+// After parsing, we need to re-escape HTML so we don't remove literal <> and &
+// For example if the original text was "The &lt;pre&gt; element is an HTML element", we want
+// our final output to be "The &lt;pre&gt; element is an HTML element", not "The <pre> element
+// is an HTML element" (which would get stripped if we ran it through this again)
+// Returns a new string!
+static char* reescape_xml_entities(const char* text)
+{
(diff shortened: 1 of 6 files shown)


brendanlong:
After this, the escape entities show in the article list (but not the title):

Screenshot from 2019-07-13 14-45-40

Still need to figure that piece out. Maybe we want to run this through XML entity parsing once when we display in a GTK title.

brendanlong:
The latest commit fixes the rendering so we handle HTML escaping in title. Most likely the better solution would be to ensure we only strip HTML once, but that may be overly complicated.

brendanlong:
Final version:

Screenshot from 2019-07-13 15-09-59

PR #934, opened 2019-08-15
tt-rss: fix feeds order

yann-soubeyrand:
feeds seems to be a linked list. Add operation on linked list is often done by adding the element at the head of the list, not the tail. I did not find information on this specific implementation (https://valadoc.org/gee-0.8/Gee.LinkedList.add.html), but the observed behaviour is addition to the head. However, I can't be sure if this an API contract or an implementation detail.

The consequence of this behaviour is that the feeds appear in reverse order in the interface. This PR should fix it.

### DIFF
--- a/plugins/backend/ttrss/ttrssAPI.vala
+++ b/plugins/backend/ttrss/ttrssAPI.vala
@@ -222,7 +222,7 @@ public class FeedReader.ttrssAPI : GLib.Object {
 
 					for(uint i = 0; i < feed_count; i++)
 					{
-						var feed_node = response.get_object_element(i);
+						var feed_node = response.get_object_element(feed_count - i - 1);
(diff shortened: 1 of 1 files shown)


plugins/backend/ttrss/ttrssAPI.vala
```diff
@@ -222,7 +222,7 @@ public class FeedReader.ttrssAPI : GLib.Object {
 
 					for(uint i = 0; i < feed_count; i++)
```
bilelmoussaoui: Why not just do the for starting from feed_count -1?

yann-soubeyrand: I first wrote
```
for(uint i = (feed_count - 1); i >= 0; i--)
```
then I realized i was unsigned. Another solution would be to type i as an int instead of an uint. Matter of taste I guess ;-)

PR #937, opened 2019-08-24
Drop Keywords from desktop files

jtojnar:
1. `X-GNOME-Keywords` was surpassed by `Keywords` key, added in the version 1.1 of the XDG desktop-entry-spec.

2. It is empty so there is no need to have it there.

3. According to the spec, “Trailing empty strings must always be terminated with a semicolon.” Since this is not the case, gettext makes it eat the supported types from the following line:

```
X-GNOME-Keywords=MimeType=x-scheme-handler/feedreader;
```


bilelmoussaoui:
Thanks!

PR #938, opened 2019-09-01
typo family, added

leela52452:
is it OK to replace other typos in this file


jangernert:
@leela52452 sure, correct whatever typos you find. It's greatly appreciated.

Issue #933, opened 2019-08-14
[UX] "Content Grabber" not adequately explained in Settings


trwnh:
In Settings > Internals, there is a toggle labeled "Content Grabber". It is not immediately clear what this means. Judging by the context clue that the other two options under "Additional features" were labeled "Download images" and "Internal media player", I assumed it had something to do with caching media and attachments. It was only after I searched through this Github repo that I discovered it was instead a way to fetch the full article and expand "read more" links -- something that I only discovered because I was about to file a feature request for a full-text fetcher/expander to be added, ironically.

I propose either renaming the "Content Grabber" feature to be more explicit (e.g. "Article Expansion") or otherwise to add subtext (e.g "Fetch full text of web pages"). Perhaps subtext could be added to other options as well, for that matter.

bilelmoussaoui:
Mind sending a MR to improve the wording?

trwnh:
I'm not sure how GTK works, so the most I could do right now would be to either rename the text label or to provide a mockup of what subtext might look like. Is there any feedback to which phrasing option might be best or most preferred?

bilelmoussaoui:
Just changing the labels for now would be enough to help the future user understand what the future does. We can improve things later. You can take a look at for some help with the phrasing style https://developer.gnome.org/hig/stable/writing-style.html.en

trwnh:
Do I need to change it only in one file, or in multiple files? I'm not sure where I should make the change(s).

bilelmoussaoui:
https://github.com/jangernert/FeedReader/blob/master/src/Widgets/SettingsDialog.vala#L162 you just need to change `_("Content Grabber")` I will take care of the rest.

PR #941, opened 2019-09-07
Rename "Content Grabber" in Settings

trwnh:
Per #933

This should probably be more clear while remaining short. I still think it could use subtext to explain that this is about the "read more" truncation, but this is a first step https://github.com/jangernert/FeedReader/issues/933#issuecomment-529095213


bilelmoussaoui:
Thanks!

PR #943, opened 2019-09-13
Correct secret collection flag

bilelmoussaoui:
Backported patch from Flathub build with GNOME 3.34 runtime


bilelmoussaoui:
We probably need to bump the image to fedora 31 or probably just switch to Flatpak builds only :p

jangernert:
@bilelmoussaoui Any idea why the circleci build fails?

bilelmoussaoui:
We need to update the docker image to get a newer Vala release

PR #948, opened 2019-09-26
Update libdecsync

39aldo39:
Update libdecsync and adjust the plugin as well.

The new libdecsync version adds a compatibility check for future versions. I am also working on creating a shared library for libdecsync, but it may be more work to change to it. This update should at least ensure compatibility.


jangernert:
Thanks :)

PR #956, opened 2019-10-23
fix: freshrss with yunohost-like solutions

tytan652:
With self-hosted solution like YunoHost, FreshRSS is behind a SSO/LDAP solution then a ping with the base url will give a 404 error.
But by using the api url (like https://demo.freshrss.org/api/) we can prove that the server is reachable with that type of solution because base_url+"/api/" and base_url+"/api/greader.php" are "whitelisted"

PS: a ping with base_url+"/api/greader.php" provoke a http error 400 Bad Request

### DIFF
--- a/plugins/backend/fresh/fresh.plugin
+++ b/plugins/backend/fresh/fresh.plugin
@@ -2,7 +2,7 @@
 Module=fresh
 Loader=C
 Name=freshRSS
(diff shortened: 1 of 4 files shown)


tytan652:
Error in the new function getApiURL() correction in progress

PR #962, opened 2019-11-13
Don't add feeds on the UI thread

jameysharp:
On at least the DecSync backend, `addFeed` involves downloading the specified feed, and that's slow. Doing it synchronously means that the UI hangs for a while. I've never written any Vala before but this patch still lets me add feeds without stopping the UI, so hopefully it's good?


jangernert:
thanks!

PR #977, opened 2020-01-12
get_pdo -> get_dbh

nielsk:
After changing this the plugin works again with tt-rss and doesn't kill it.


jangernert:
Thank you for the fix.

PR #982, opened 2020-02-10
Logger: Don’t set a custom log writer function

### LINKED ISSUE #970: FeedReader causes excessive disk writes
I noticed that gnome-shell is writing gigabytes of data to disk. Checking out SystemD logs for gnome-shell, it turns out that out of 1001 entries in the logs for gnome-shell, 974 are caused by FeedReader.

gnome-shell and systemd are both reported to have written gigabytes of data to disk, 7.6 GB and 14 GB, respectively. Next in the top disk writers are gvfs and Firefox, each around 400 MB and both expected to write data. gnome-shell is not expected to write data at all.

Something is not right here. For now it looks like FeedReader is logging way too much stuff, especially uninteresting stuff. I will try to investigate it more.

pwithnall:
Unless you’re going to reimplement all the functionality from
g_log_writer_default() (or deliberately don’t want it), you should not
set a custom writer function. g_log_writer_default() likely does what
you want.

In particular, setting the writer function to
g_log_writer_standard_streams() meant that `G_MESSAGES_DEBUG` was
ignored, resulting in all debug messages being logged unconditionally
and sending gigabytes of data to the disk over the course of several
days. This noticeably slowed down people’s systems.

It also meant that fatal messages did not actually abort the program,
which hides bugs.

g_log_writer_default() handles those things, then tries sending log
messages to the journal. If that fails, it actually calls
g_log_writer_standard_streams().

Signed-off-by: Philip Withnall <withnall@endlessm.com>

Fixes: https://github.com/jangernert/FeedReader/issues/970

### DIFF
--- a/src/Logger.vala
+++ b/src/Logger.vala
@@ -49,6 +49,5 @@ public class FeedReader.Logger : GLib.Object {
 	public static void init(bool verbose)
 	{
 		m_log_debug_information = verbose;
-		GLib.Log.set_writer_func((LogWriterFunc)GLib.Log.writer_standard_streams);
 	}
 }


pwithnall:
Would you mind doing a micro release of FeedReader to contain this fix please? The bug is causing a lot of disk and power consumption for every user of FeedReader.

